PR #1, opened 2019-06-03
[cookie-cut] Initial cookie-cut

wlonk:
biv


wlonk:
This is going from zero to something, so it's huge. Probably best to evaluate spots you're curious about, and mostly just try to get it set up per the Readme and note friction or errors.

pytest.ini
```diff
@@ -0,0 +1,9 @@
+[pytest]
+python_files = *.py
+norecursedirs = .* _* node node_modules coverage venv
```
wlonk: I'm thinking about restructuring the tests here (and eventually in the template) to be more normal, and more easily excluded from the Django code tree. Like:

```
┬ metashare
│ └ foo
│    └ bar.py
└ tests
  └ foo
     └ test_bar.py
```

Thoughts?

pytest.ini
```diff
@@ -0,0 +1,9 @@
...
+addopts =
+    --tb short
+    --cov=metashare
+    --cov-report html
+    --cov-report term
```
wlonk: Half the time I have to remove this line while working on a subset of tests or a known-failing test, as it causes the coverage report to be so long that it pushes the failure into the scrollback. It's useful when I expect the tests to all pass and want to see that 100%, but I might remove it by default here.

pytest.ini
davisagli: Any particular motivating reason? I haven't had trouble with the current layout so I'm not sure it's worth the effort to change.

wlonk: Two reasons: a change in my philosophy on the question (which would not be worth changing here) and the fact that tests for templatetags need special-casing because of how Django detects available templatetags.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,264 @@
...
+
+    BUCKETEER_AWS_ACCESS_KEY_ID=...
+    BUCKETEER_AWS_SECRET_ACCESS_KEY=...
+    BUCKETEER_BUCKET_NAME=...
+
+3. Run ``docker-compose build`` to build/re-build all the container images.
+
+4. Run ``docker-compose up`` to start the server(s).
+
+5. Visit `<http://localhost:8080/>`_ in your browser.
```
jgerigmeyer: @wlonk This currently isn't working for me -- maybe a port forwarding problem?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,264 @@
...
+To run the local development server::
+
+    yarn serve
+
+The running server will be available at `<http://localhost:8080/>`_.
+
+.. _Redis Quick Start: https://redis.io/topics/quickstart
+
+Logging in with Salesforce
+~~~~~~~~~~~~~~~~~~~~~~~~~~
```
jgerigmeyer: I think this should probably be copied above to the Docker setup too, with commands edited to work through Docker.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,264 @@
...
+In commit messages or pull request titles, we use the following emojis to label
+which development commands need to be run before serving locally (these are
+automatically prepended to commit messages):
+
+- 📦 (``:package:``) -> ``pipenv install --dev``
+- 🛢 (``:oil_drum:``) -> ``python manage.py migrate``
+- 🧶 (``:yarn:``) -> ``yarn``
+
+Internationalization
+~~~~~~~~~~~~~~~~~~~~
```
jgerigmeyer: This section should also be copied above with Docker-specific commands.

CONTRIBUTING.rst
```diff
@@ -0,0 +1,264 @@
...
+
+To build and compile ``.mo`` and ``.po`` files for the backend, run::
+
+   $ python manage.py makemessages --locale <locale>
+   $ python manage.py compilemessages
+
+These commands require the `GNU gettext toolset`_ (``brew install gettext``).
```
jgerigmeyer: Adjust this to work via Docker.

package.json
```diff
@@ -0,0 +1,150 @@
...
+    "pytest": "pytest",
+    "pytest:report-coverage": "$VIRTUAL_ENV/bin/coveralls",
+    "pytest:check-coverage": "coverage3 report --fail-under=100",
+    "build": "webpack --config webpack.dev.js",
+    "prod": "webpack --config webpack.prod.js",
+    "heroku-postbuild": "yarn prod"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged",
```
jgerigmeyer: This seems to be overridden by the Docker-specific pre-commit hook, so it breaks if Docker is not setup.

CONTRIBUTING.rst
wlonk: Yep, I'm building and working on that right now.

package.json
wlonk: Hm, since the `githooks` Docker changes the surrounding dev env, I am not sure what the best way around that might be. I'll consider it.

jgerigmeyer: Maybe we should discuss -- I'm not clear what the dedicated `githooks` container is doing. It seems worst case we can ditch lint-staged and just modify the Docker githook to check whether Docker is running?

wlonk: Let's discuss, yeah. It's doing something very different from what I think you think it's doing.

jgerigmeyer: I don't think I'm far off, but thanks. :) It's copying the hook from `hooks` into the local `.git/hooks` directory, no? Which replaces the husky `pre-commit` hook...

In which case I guess this is only a problem for users who have already run the Docker container... so let's just leave it alone. 👍

wlonk: Right; that container has to run exactly once, and it modifies the host environment. So when you said "check whether Docker is running" I thought you thought it was more of a continuous process. But yeah, it's just a one-off "get this hook symlinked" thing, if you run the Docker stuff. If you never do, it'll remain untouched.

webpack.dev.js
```diff
@@ -25,12 +25,14 @@ module.exports = merge(common, {
...
     proxy: {
-      '**': 'http://localhost:8000',
+      '**': 'http://backend:8000',
       '/ws': {
-        target: 'http://localhost:8000',
+        target: 'http://backend:8000',
         ws: true,
       },
     },
+    host: '0.0.0.0',
```
jgerigmeyer: This will break non-Docker development. Maybe we can if-branch this setting based on an env var or something.

app.json
```diff
@@ -0,0 +1,88 @@
+{
+  "name": "MetaShare",
+  "description": "A tagline to end all taglines.",
```
davisagli:
```suggestion
  "description": "Web-based tool for collaborating on Salesforce projects",
```

bin/unpack-node
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash -e
+
+mkdir -p node
+pushd node
+
+config=`../bin/config.guess`
```
davisagli: Maybe check Python's `sys.platform` instead of bundling config.guess?

metashare/logfmt.py
```diff
@@ -0,0 +1,103 @@
...
+            f"tag={tag}",
+            f"module={record.module}",
+        ]
+        if parsed_msg:
+            for k, v in msg.items():
+                fields.append(f"{k}={v}")
+        else:
+            fields.append(f"msg={msg}")
+        fields.append(f"{rest}")
+        return " ".join(filter(None, fields))
```
davisagli: This should really go in sfdo-template-helpers rather than maintaining multiple copies.

runtime.txt
```diff
@@ -0,0 +1 @@
+python-3.7.2
```
davisagli: 3.7.3 is latest

bin/unpack-node
jgerigmeyer: This comes back to the question of whether we want/need to support non-Docker local development (or at least node/yarn version-specific local dev). If not, this can just be removed.

app.json
wlonk: This was, clearly, a placeholder 😆

CONTRIBUTING.rst
```diff
@@ -0,0 +1,268 @@
...
+Logging in with Salesforce
+--------------------------
+
+To setup the Salesforce OAuth integration, run the ``populate_social_apps``
+management command. The values to use in place of the ``XXX`` and ``YYY`` flags
+can be found on the Connected App you've made in your Salesforce configuration,
+or if you're an OddBird, you can find these values in the shared Keybase team
+folder (``metashare/prod.db``)::
+
+    $ docker-compose run web python manage.py populate_social_apps --prod-id XXX --prod-secret YYY
```
jgerigmeyer: This command fails for me:

```
  File "/app/metashare/multisalesforce/management/commands/populate_social_apps.py", line 37, in handle
    self.create_production_app(options["prod_id"], options["prod_secret"])
  File "/app/metashare/multisalesforce/management/commands/populate_social_apps.py", line 27, in create_production_app
    self._create_app("production", "https://login.salesforce.com/", id, secret)
  File "/app/metashare/multisalesforce/management/commands/populate_social_apps.py", line 21, in _create_app
    name=f"Salesforce {name.title()}", key=key, client_id=id, secret=secret
  File "/usr/local/lib/python3.7/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 538, in get_or_create
    return self.get(**kwargs), False
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 402, in get
    num = len(clone)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 256, in __len__
    self._fetch_all()
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 1242, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 55, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
  File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1100, in execute_sql
    cursor.execute(sql, params)
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 99, in execute
    return super().execute(sql, params)
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 76, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.7/site-packages/django/db/utils.py", line 89, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/usr/local/lib/python3.7/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python3.7/site-packages/newrelic/hooks/database_psycopg2.py", line 51, in execute
    **kwargs)
  File "/usr/local/lib/python3.7/site-packages/newrelic/hooks/database_dbapi2.py", line 25, in execute
    *args, **kwargs)
django.db.utils.ProgrammingError: relation "socialaccount_socialapp" does not exist
LINE 1: ...p"."secret", "socialaccount_socialapp"."key" FROM "socialacc...
```

wlonk: Mm, we're lacking some apps I think. I'll get on it.

davisagli: Is it that, or not having run the initial migration?

wlonk: `start-dev.sh`, which the `web` Dockerfile runs, runs migrations.

davisagli:
It built and started up on the first try. What wizardry is this.

I haven't tried setting up the social auth because I'm anticipating that it will need adjustment for this app what with authing to both github and SF.

wlonk:
@davisagli Yay, and also yes, it will need adjustment.

bin/unpack-node
wlonk: I think we can remove this, given today's decisions.

jgerigmeyer: @wlonk This will still be used by CI to ensure the right node/yarn versions, unless/until we switch to Docker there.

wlonk: OK 👍

compose/client/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:10.16.0
```
jgerigmeyer: @wlonk Would it be possible to pull this from `package.json` instead of duplicating it here?

CONTRIBUTING.rst
```diff
@@ -0,0 +1,268 @@
...
+(If you are using local development, omit the ``docker-compose run web``.)
+
+You can also run it with ``--test-id`` and ``--test-secret``, or ``--cust-id``
+and ``--cust-secret``, or all three sets at once, to populate all three
+providers.
+
+Once you've logged in, you probably want to make your user a superuser. You can
+do that easily via the ``promote_superuser`` management command::
+
+    $ docker-compose run web python manage.py promote_superuser <your email>
```
jgerigmeyer: @wlonk Accessing the Django admin seems to be broken in Docker-dev:

```
Environment:


Request Method: GET
Request URL: http://localhost:8080/admin

Django Version: 2.2.2
Python Version: 3.7.3
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.sites',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'channels',
 'whitenoise.runserver_nostatic',
 'django.contrib.staticfiles',
 'django_rq',
 'scheduler',
 'allauth',
 'allauth.account',
 'allauth.socialaccount',
 'rest_framework',
 'rest_framework.authtoken',
 'parler',
 'metashare',
 'metashare.multisalesforce',
 'metashare.api',
 'metashare.adminapi.apps.AdminapiConfig',
 'django_js_reverse',
 'django_extensions']
Installed Middleware:
['metashare.logging_middleware.LoggingMiddleware',
 'sfdo_template_helpers.admin.middleware.AdminRestrictMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware']



Traceback:

File "/usr/local/lib/python3.7/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/src/sfdo-template-helpers/sfdo_template_helpers/admin/middleware.py" in __call__
  27.                 self._validate_ip(request)

File "/src/sfdo-template-helpers/sfdo_template_helpers/admin/middleware.py" in _validate_ip
  36.             raise SuspiciousOperation(f"Disallowed IP address: {ip_addr}")

Exception Type: SuspiciousOperation at /admin
Exception Value: Disallowed IP address: 172.31.0.2
```

compose/client/Dockerfile
wlonk: I'm not sure how we might. We could set the value at `Dockerfile`/`package.json` creation time, but I think getting the value at runtime is gonna be tricky.

If we set up a makefile, it could have a "copy the value in to the Dockerfile" step before build?

jgerigmeyer: Not a big deal, just curious 👍

CONTRIBUTING.rst
wlonk: Hah! So, this doesn't happen in local dev because we're proxying from `localhost` to `localhost`. Here, the two Docker containers have their own IPs. I'll see about fixing it, but in the meantime, you can access it from `http://localhost:8000/admin/`.

jgerigmeyer: Same error at :8000.

wlonk: Oh, yeah. This is because of the SFDO "if it's not localhost, make sure it's on our subnet" middleware. :-/ I'm working on a fix.

davisagli: @wlonk you can set the default allowed subnets to be inclusive. we'll set it appropriately in production.

wlonk:
1. Sounds good.
2. Expected; you have to include the trailing slash, otherwise the fall-through to the frontend takes over. See similar behavior on MetaDeploy.
3. Possibly when you rebuild; perhaps we should caution against rebuilding `postgres`, and suggest only rebuilding `web` and `client`? If that hypothesis is right, anyway.
4. I have noticed this too, and haven't yet figured out why it happens.

jgerigmeyer:
>     1. Sounds good.
> 
>     2. Expected; you have to include the trailing slash, otherwise the fall-through to the frontend takes over. See similar behavior on MetaDeploy.
> 
>     3. Possibly when you rebuild; perhaps we should caution against rebuilding `postgres`, and suggest only rebuilding `web` and `client`? If that hypothesis is right, anyway.
> 
>     4. I have noticed this too, and haven't yet figured out why it happens.

1. Done.
2. 👍
3. Okay, though that seems like a last-resort and not ideal. Could you test the hypothesis? If true, is there any way we could warn or require a prompt before allowing a rebuild of postgres? Even if we add a caution, it seems likely someone could do `docker-compose build` without realizing it.
4. 👌

wlonk:
@jgerigmeyer lemme test the `postgres` rebuild hypothesis first.

wlonk:
@jgerigmeyer OK, that hypothesis is dead, so I will happily merge this and revisit why your db is vanishing later, if that's OK!

PR #3, opened 2019-06-07
[single-docker] Build server and client in same Docker image

wlonk:
This more closely mirrors how we do things on Heroku, and may provide a path to simply deploying the appropriate Dockerfile on Heroku. It may also simplify VS Code intellisense setup.

Related to https://trello.com/c/RXQLKT3b/46-i-want-to-use-docker-for-deployment-and-ci

catte


package.json
```diff
@@ -88,12 +88,12 @@
...
   "scripts": {
     "webpack:serve": "webpack-dev-server --config webpack.dev.js",
-    "django:serve": "python manage.py runserver",
+    "django:serve": "python manage.py runserver 0.0.0.0:8000",
     "redis:clear": "redis-cli FLUSHALL",
     "worker:serve": "python manage.py rqworker default short",
     "scheduler:serve": "python manage.py rqscheduler default short",
     "rq:serve": "npm-run-all redis:clear -p worker:serve scheduler:serve",
-    "serve": "run-p django:serve webpack:serve rq:serve",
+    "serve": "run-p django:serve webpack:serve",
```
jgerigmeyer: @wlonk Can you say more about this change?

package.json
```diff
@@ -88,12 +88,12 @@
   },
   "scripts": {
     "webpack:serve": "webpack-dev-server --config webpack.dev.js",
-    "django:serve": "python manage.py runserver",
+    "django:serve": "python manage.py runserver 0.0.0.0:8000",
```
jgerigmeyer: I'm guessing this will break local non-Docker dev?

package.json
wlonk: I wanted to actually flag it before you reviewed, but you beat me to it. This is just a quick change of convenience, intended to get the Docker image running; the worker stuff was erroring, so I just took it out for now. This is not intended to be the going-forward way.

package.json
wlonk: Again, a quick change of convenience. We can easily get around these two things breaking things by making distinct Docker yarn scripts.

wlonk: So, `0.0.0.0` will bind to localhost as well as outside interfaces; binding to `0.0.0.0` won't break things. But it will make them brittle, as we're inlining the port there.

PR #7, opened 2019-06-11
[oddbird-base-image] Move base image stuff into base image

wlonk:
whale


jgerigmeyer:
This would simplify CI (CircleCI could just use the published base image).

It's unfortunate to have Node version locked in the image, but I guess that's unavoidable -- and we could always update within the image or update the image.

jgerigmeyer:
This would simplify CI (CircleCI could just use the published base image).

It's unfortunate to have Node version locked in the image, but I guess that's unavoidable -- and we could always update within the image or update the image.

Seems like it might make more sense to have this owned by SFDO-Tooling or the like?

davisagli:
@jgerigmeyer would it affect the runtime of CI or just simplicity of the configuration file? Because not depending on a separate image is simpler in a different way.

jgerigmeyer:
@davisagli I'm noting it as a potentially positive side-effect, not the primary reason to make this change -- I'd like @wlonk to speak to that.

It would simplify both runtime and configuration. On CircleCI (unless we want to switch to HerokuCI), we basically have three choices (if we want to use our Docker setup):

1. Run directly on a Linux VM and spin up our Docker image (slower and not preferred, but provides more control.
2. Run in a base Docker image (e.g. circleci/python-3.7) and use a "Remote Docker Environment" to spin up our Docker image.
3. Run directly in our custom Docker image, which requires it to be published to Dockerhub.

Cf. https://circleci.com/docs/2.0/executor-types/ and https://circleci.com/docs/2.0/building-docker-images/

davisagli:
@jgerigmeyer I guess my main concerns are:
- I'd like to be able to monitor changes to the image. No problem, I just watched its repo.
- I'd like to be able to fork the repo that the image is built from if it is modified for other OddBird projects and diverges from our needs at some point. No problem, as long as @wlonk doesn't rage quit github that should be possible.

So I guess I'm fine experimenting with this.

wlonk:
I pushed this without explaining my thought process, but it looks like y'all divined much of it.

I set this up in my personal Docker Hub and GitHub accounts just as a way to experiment with it. The intent would be to make an `oddbird` base image in our own namespaces, and probably to fork it for SFDO purposes.

Regarding baking in the Node and Python versions, the intent is to break this out into a template, and be able to build different tagged versions, but that's more work down the road, so I figured baking it in would get us enough for the experiment more quickly.

Once both of those are the case, we would have a `FROM` line like `FROM sfdo/base:py3.7-node10.16`.

Does that all sound good and clear?

PR #9, opened 2019-06-12
[all-docker-all-the-time] Use our base Docker image in CI

wlonk:
sea lion

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
           name: Install Dependencies
           command: |
             yarn
-            mkdir venv
+            python3 -m venv venv 
             pipenv install --dev
       - run:
           name: Check that yarn.lock matches package.json
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,6 @@ jobs:
       - run:
           name: Lint
           command: |
-            pipenv shell
             yarn lint
 
   test-client:
@@ -111,12 +110,12 @@ jobs:
       - run:
           name: Run Database Migrations
           command: |
-            pipenv shell
+            . venv/bin/activate
             python manage.py migrate --no-input
       - run:
           name: Test Python
           command: |
-            pipenv shell
+            . venv/bin/activate
             pytest --junitxml=pycov/pytest-junit.xml
       - store_artifacts:
           path: htmlcov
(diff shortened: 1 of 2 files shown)


wlonk:
This uses our base image, but it doesn't use _project's_ image. Maybe it should? Still a WIP.

.circleci/config.yml
```diff
@@ -63,14 +66,18 @@ jobs:
...
+      VIRTUAL_ENV: ./venv
     steps:
       - checkout
       - attach_workspace:
           at: .
       - run:
           name: Test JS
-          command: yarn test:coverage
+          command: |
+            . venv/bin/activate
```
jgerigmeyer: Why does the venv need to be active to run `yarn`?

.circleci/config.yml
```diff
@@ -35,8 +36,8 @@ jobs:
       - persist_to_workspace:
           root: .
           paths:
-            - ./venv
-            - ./node_modules
+            - venv
+            - node_modules
```
jgerigmeyer: Are these necessary changes, or side-effects from lots of small tweaks trying to get CI to pass?

.circleci/config.yml
```diff
@@ -16,9 +18,8 @@ jobs:
           name: Install Dependencies
           command: |
             yarn
-            python3 -m venv venv
+            python3 -m venv venv 
             . venv/bin/activate
-            pip install pipenv
             pipenv install --dev
```
jgerigmeyer: Should we add `--ignore-pipfile` here to match the project Dockerfile?

wlonk: Yeah, I removed it while trying to debug.

.circleci/config.yml
wlonk: The latter.

.circleci/config.yml
wlonk: It doesn't, here. This was just a side-effect of thrashing.

.circleci/config.yml
```diff
@@ -3,7 +3,9 @@ jobs:
   build:
     working_directory: ~/metashare
     docker:
-      - image: circleci/python:3.7.3-node
+      - image: wlonk/oddbird
+    environment:
+      VIRTUAL_ENV: ./venv
```
jgerigmeyer: @wlonk What do you think about setting this in CircleCI instead of repeating it here?

wlonk: Unless the value is secret, I much prefer setting it in the config, so it can be in source control and tracked.

PR #10, opened 2019-06-14
[typescript] Flow -> Typescript

jgerigmeyer:
cat typing


CONTRIBUTING.rst
```diff
@@ -92,6 +92,9 @@ the Docker container. If you do not see the prompt, run the "Remote-Containers:
 Open Folder in Container..." command from the VS Code Command Palette to start
 the Docker container.
 
+The first build will take a number of minutes, but subsequent builds will be
+significantly faster.
```
wlonk:
```suggestion
significantly faster. Docker caches each command in the `Dockerfile` as its own layer, so if
you change the `Dockerfile`, changing earlier layers will bust the cache on the lower layers,
and make your next build slow again.
```

.devcontainer.json
```diff
@@ -37,10 +38,13 @@
       "javascript",
       "javascriptreact",
       {
-        "language": "vue",
+        "language": "typescript",
         "autoFix": true
       },
-      "typescript"
+      {
+        "language": "typescriptreact",
```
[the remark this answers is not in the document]
jgerigmeyer: 👍

.devcontainer.json
```diff
@@ -51,40 +55,35 @@
     "python.pythonPath": "/usr/local/bin/python",
     "restructuredtext.linter.extraArgs": ["--max-line-length 80"],
     "[javascript]": {
-      "editor.formatOnSave": true,
+      "editor.formatOnSave": false,
```
davisagli: @jgerigmeyer why no more format on save?

jgerigmeyer: I was playing with `eslint.autoFixOnSave: true` above, which replaces this... but I'm not sure I like it. I'll probably switch it back.

PR #14, opened 2019-06-17
[favicon] wordmark favicon

stacyk:
Trello: https://trello.com/c/sTH67mNe/8-i-want-a-metashare-wordmark

Included are the generated files. The site manifest would need to be updates with the relative URLs if we are using that file. I looked at how metadeploy was set up and it is using some registered globals that don't yet exist (I think) in MetaShare yet. I've only added the files as of now, but we will need to link to them.


jgerigmeyer:
We didn't worry about a manifest.json file on MD -- I'd be inclined to make the same call here (at least for now)?

stacyk:
> We didn't worry about a manifest.json file on MD -- I'd be inclined to make the same call here (at least for now)?
I believe it is mostly for Android and Windows Phone users so likely not much of an issue here unless someone is using an android tablet and want to save this as a shortcut. Probably a super small if any market.

jgerigmeyer:
[I want a MetaShare wordmark](https://trello.com/c/sTH67mNe/8-i-want-a-metashare-wordmark)

PR #23, opened 2019-06-24
Docker deploy

wlonk:
Deploy via Docker.

See example deployment at https://salty-oasis-85874.herokuapp.com/

See https://trello.com/c/RXQLKT3b/46-i-want-to-use-docker-for-deployment-and-ci

rhino


Dockerfile
```diff
@@ -18,7 +24,11 @@ RUN pipenv install --dev --system --deploy --ignore-pipfile
...
 COPY ./yarn.lock /app/yarn.lock
 
+COPY . /app
+
 # Actually run things:
 WORKDIR /app
 RUN yarn install
+RUN yarn prod
+RUN python manage.py collectstatic --noinput
 CMD /start-dev.sh
```
davisagli: should rename this as it's not just for dev

Dockerfile
```diff
@@ -18,7 +24,17 @@ RUN pipenv install --dev --system --deploy --ignore-pipfile
 COPY ./package.json /app/package.json
 COPY ./yarn.lock /app/yarn.lock
 
-# Actually run things:
+COPY . /app
```
jgerigmeyer: Can you clarify why this is added? Does it mean the above lines (copying package.json and yarn.lock) are no longer needed? Does it copy _everything_ in the local directory, including ignored files/directories?

compose/web/start-server.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+# Can this be in the Dockerfile?
```
jgerigmeyer: Good question. Can it? 😄

docker-compose.yml
```diff
@@ -15,7 +15,8 @@ services:
       - redis
     env_file: .env
     environment:
-      - REDIS_URL=redis://redis:6379
+      DJANGO_SETTINGS_MODULE: config.settings.local
```
jgerigmeyer: Interesting; why this here? Is docker-compose not run on Heroku?

wlonk:
@jgerigmeyer your questions all interrelate, so I can't answer them each in-thread, but will address them here:

`docker-compose` _approximately_ lets you set some command line values in the yaml. We set volumes and env variables to, _approximately_, override the values in the `Dockerfile`. But Heroku doesn't do `docker-compose` at all. It does an image per proc-type, with its own context for wiring them together and providing persistent resources like DBs and queues. So we have to be able to build in the source we'll run into the Docker image, since the way the container runtime works is that we actually build the image locally, upload it to Heroku's container registry, then Heroku runs it with a bunch of `-e` flags setting appropriate environment variables.

And that's why we can't actually move the `migrations` call into the image-build; it doesn't build in the Heroku context, it just runs there. So we don't know the DB (I mean, we can fake it, but it would entwine steps that maybe shouldn't be entwined; migrating a real DB from a local Docker build seems bad). But maybe we make a `release` proc-type and put migration in there, and don't have it in the spin-up.

wlonk:
(We may be able to get some traction out of https://devcenter.heroku.com/articles/build-docker-images-heroku-yml though.)

compose/web/start-server.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# Use the presence of $PORT as a proxy for "are we local or Heroku?"
+if [ -z ${PORT+x} ];
+then
+    # PORT unset, we presume this is local dev:
+    PORT=8000 yarn serve
```
jgerigmeyer: run migrations here?

wlonk:
We can build it on Heroku now, automatically from merges to `master` on GitHub. I think that the settings in the `Dockerfile` are OK as they are, but we can continue to experiment with them moving forward.

wlonk:
(And the migrations-on-release works on Heroku, and on-start works locally.)

jgerigmeyer:
@wlonk Great! I think there was still some cleanup you wanted to do with a heroku.yml file?

wlonk:
@jgerigmeyer those changes are in the `heroku.yml` already. This is ready to review; it should be able to build on Heroku when merges are made into `master`.

PR #32, opened 2019-06-24
[product-list] Add products list view.

wlonk:
https://trello.com/c/Xia3JFsW/11-i-want-to-select-from-a-list-of-products-that-i-have-contributor-access-to-on-github

pingling


jgerigmeyer:
[I want to select from a list of products that I have contributor access to on github](https://trello.com/c/Xia3JFsW/11-i-want-to-select-from-a-list-of-products-that-i-have-contributor-access-to-on-github)

metashare/api/views.py
```diff
@@ -17,3 +18,8 @@ def get_queryset(self):
 
     def get_object(self):
         return self.get_queryset().get()
+
+
+class ProductViewSet(viewsets.ModelViewSet):
+    queryset = Product.objects.all()
+    serializer_class = ProductSerializer
```
jgerigmeyer: @wlonk This endpoint should only return products that are in the DB _and that the logged-in user has contributor access to on GitHub_. I don't see anything related to the latter requirement in this PR?

metashare/api/models.py
```diff
@@ -90,3 +95,14 @@ def valid_token_for(self):
         if all(self.token) and self.org_id:
             return self.org_id
         return None
+
+
+class Product(HashIdMixin):
+    name = models.CharField(max_length=50, unique=True)
+    repo_name = models.SlugField(unique=True)
```
jgerigmeyer: Is this just the name, or also the org/owner? Or is it the repo URL?

wlonk: It's the repo _name_ on GitHub, which given the talk about having everything be under one org, can be used to get the full URL.

jgerigmeyer: Makes sense. I assume we'll want that org setting controlled by an env var or something, not hard-coded.

metashare/api/views.py
jgerigmeyer: There's a distinction between access to _create new products_ and access to view and contribute to products, I think? As far as I'm aware the details haven't been nailed down yet, but contributor access seems like the place to start for this story at least.

metashare/api/models.py
davisagli: It's likely that we will only want to configure one GitHub organization, but I don't think we want to build that assumption into the data model. I think we want to store both org and repo name and have a URL structure like /products/[org]/[repo]

metashare/api/models.py
```diff
@@ -90,3 +94,24 @@ def valid_token_for(self):
...
+    version_number = models.CharField(max_length=50)
```
davisagli: I guess this is here because it's in the mockup, but I don't think this is very relevant for MetaShare, which is focused on development, not releasing versions. Showing a last modified date would probably be more useful.

metashare/api/models.py
```diff
@@ -90,3 +94,24 @@ def valid_token_for(self):
...
+    description = MarkdownField()
+    is_managed = models.BooleanField(default=False)
+    license = ArrayField(
+        models.CharField(max_length=64, choices=LICENSES), blank=True, default=list
+    )
```
davisagli: Not sure this needs to be a field on the model; it's only going to be used as input when generating a new product's repository, right?

Also making it a list seems weird (_especially_ with a singular field name). Including multiple licenses in a repository is a bit of an edge case and it's fine if we have to tell people to edit files manually for that case.

metashare/api/models.py
```diff
@@ -90,3 +94,24 @@ def valid_token_for(self):
...
+
+
+class ProductDependency(HashIdMixin):
+    url = models.URLField()
+    product = models.ForeignKey(
+        Product, on_delete=models.CASCADE, related_name="dependencies"
+    )
+
+    def __str__(self):
+        return self.url
```
davisagli: Dependencies should work like this:

Products should know their own github URL and should have a checkbox for whether the product is available to choose as a dependency when adding a new product.

Let me know if I'm forgetting a reason we need it, but I don't think we need/want the relationship between a product and its dependencies to be part of the MetaShare data model. The truth for this is the list of dependencies in cumulusci.yml, and I don't want us to have to worry about keeping this in sync with that unless there's a reason.

wlonk: OK; I began work on `product-list`, but drew from the descriptions in the `product-create` story; I understand now that a number of those things are not things we need to store in the DB, but rather take in as input for when we initialize a new product repo. Gotcha.

docs/api/products.rst
```diff
@@ -0,0 +1,27 @@
...
+   [
+      {
+        "id": "3Lw7OwK",
+        "name": "Test Product",
+        "repo_name": "test-product",
+        "version_number": "v0.1.0",
+        "description": "<p>This is safely rendered markdown</p>",
+        "is_managed": false,
+      }
+   ]
```
jgerigmeyer: @wlonk I think we might as well paginate these right from the beginning. Similar to MD, but no need for the `first_page` or category meshugas.

davisagli: Is pagination trivial? If not, leave it for later.

jgerigmeyer: Because we're copying over some patterns from MD, I'd say on the front end it's trivial (arguably easier than ripping it out). I'll let @wlonk speak to the back end.

metashare/api/admin.py
```diff
@@ -1,9 +1,15 @@
...
 @admin.register(User)
 class UserAdmin(admin.ModelAdmin):
     list_display = ("username", "is_active", "is_staff", "is_superuser", "date_joined")
     search_fields = ("username",)
+
+
+@admin.register(Product)
+class ProductAdmin(admin.ModelAdmin):
+    list_display = ("name", "description")
+    prepopulated_fields = {"repo_url": ("name",)}
```
jgerigmeyer: I'm not sure this still makes sense, now that it's a URL.

wlonk: Right, it doesn't.

metashare/api/views.py
```diff
@@ -1,19 +1,53 @@
...
+        user = self.get_object()
+        repos = gh.get_all_org_repos(user)
+        GitHubRepository.objects.filter(user=user).delete()
+        GitHubRepository.objects.bulk_create(
+            [GitHubRepository(user=user, url=repo) for repo in repos]
+        )
+        serializer = FullUserSerializer(user)
+        # TODO: this _should_ return the new list of products for the user, but that
+        # feels wrong for this endpoint. Merits thought.
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
jgerigmeyer: @wlonk I'm curious how you anticipate this view being used.

wlonk: I don't really know; I just felt uneasy having no way to refresh what products you can see without logging in and out again.

jgerigmeyer: Hm. So it won't be refreshed when the `products` api is requested? We're only doing that once per page-load, so I would expect it to be up-to-date...

wlonk: That's what I was saying about not wanting HTTP requests inside our HTTP requests; I compromised (with myself) by having a dedicated "this could be slow" endpoint to refresh thing.

jgerigmeyer: Hm, that doesn't seem useful from a user's perspective. I think this data should be up-to-date on a page-refresh, not just a login/logout. So how does it help to require an additional API call to refresh the data first?

wlonk: I just really really don't want to query GH each time `.../products/` is hit. It seems bad. Maybe it's not so bad, maybe there's another way. I don't know yet.

jgerigmeyer: Ok, I'm open to that. Maybe we want to add an explicit "Refresh repos" btn, similar to Pyup or Greenkeeper or Coveralls.io.

metashare/api/models.py
```diff
@@ -90,3 +94,32 @@ def valid_token_for(self):
         if all(self.token) and self.org_id:
             return self.org_id
         return None
+
+
+class Product(HashIdMixin):
+    name = models.CharField(max_length=50, unique=True)
+    repo_url = models.URLField(unique=True)
+    version_number = models.CharField(max_length=50)
+    description = MarkdownField()
```
jgerigmeyer: I think this should be optional?

metashare/api/views.py
wlonk: @stacyk do you wanna weigh in on the UX considerations there?

To recap:

We can't (we don't think) reasonably keep the API constantly in sync with GitHub's view of the world, so we have to cache data. Sometimes that cache may be stale, and we'd like to offer users a way to refresh it manually.

metashare/api/models.py
```diff
@@ -90,3 +94,32 @@ def valid_token_for(self):
         if all(self.token) and self.org_id:
             return self.org_id
         return None
+
+
+class Product(HashIdMixin):
+    name = models.CharField(max_length=50, unique=True)
+    repo_url = models.URLField(unique=True)
```
jgerigmeyer: @wlonk Do we really want `repo_url` here, or would it be easier to use `repo_owner` and `repo_name`?

If we do stick with url, we need a way to standardize it. Right now for it to work you have to enter in the exact form returned from GH, e.g. `https://api.github.com/repos/jgerigmeyer/jquery-loading-overlay`. I don't think we want to require admins to know that?

metashare/api/gh.py
```diff
@@ -0,0 +1,17 @@
...
+
+def get_all_org_repos(user):
+    try:
+        token = (
+            user.socialaccount_set.get(provider="github").socialtoken_set.get().token
+        )
+    except (ObjectDoesNotExist, MultipleObjectsReturned):
+        return []
+    gh = login(token=token)
+    return [repo.url for repo in gh.repositories()]
```
jgerigmeyer: @wlonk As far as I can tell, `repositories()` only returns repos owned by individuals, not by orgs. At least in my case I'm only seeing repos owned by myself or other individuals -- no OddBird or SFDO or other org repos. Do we need to combine this with `organizations()` and `org.repositories()` or something?

https://github3py.readthedocs.io/en/master/api-reference/github.html#github3.github.GitHub.organizations
https://github3py.readthedocs.io/en/master/api-reference/orgs.html#github3.orgs.ShortOrganization.repositories

metashare/api/models.py
jgerigmeyer: @wlonk You'll go ahead and remove/replace this?

metashare/api/serializers.py
```diff
@@ -10,3 +12,18 @@ class FullUserSerializer(serializers.ModelSerializer):
...
+    id = serializers.CharField(read_only=True)
+
+    class Meta:
+        model = Product
+        fields = (
+            "id",
+            "name",
+            "repo_url",
+            "version_number",
+            "description",
```
jgerigmeyer: Serialize this as Markdown.

metashare/api/views.py
davisagli: I think refreshing when the user logs in would be a reasonable compromise, because then there's at least a simple action we can tell users to take if they know it's not up to date.

jgerigmeyer: For comparison with how some other tools handle this. If we have the endpoint, I don't see a good reason not to add a button rather than instructions to logout/login.

Screenshot 2019-06-27 15 49 43

---

Screenshot 2019-06-27 15 50 27

---

Screenshot 2019-06-27 15 50 50

metashare/api/gh.py
wlonk: This is what I was saying in standup, that the results looked wrong; I saw some repos from orgs I'm part of, but no Oddbird ones, so I wasn't sure what was going on. I'll dig through those docs again and see.

metashare/api/models.py
davisagli: One advantage to using a full URL is that it's more flexible in case we ever want to support repositories stored somewhere other than github.com

metashare/api/views.py
davisagli: I agree, a button sounds good.

metashare/api/models.py
jgerigmeyer: Yeah, that makes sense to me. In that case I think we should:

1. ignore `.git` endings
2. convert the repos returned from GH from e.g. `https://api.github.com/repos/...` to `https://github.com/...`

metashare/api/models.py
jgerigmeyer: I removed `version` and added a task for @wlonk to add last-modified date.

docs/api/products.rst
```diff
@@ -0,0 +1,32 @@
...
+
+   HTTP/1.1 200 OK
+
+    {
+      "count": 150,
+      "next": "https://.../api/products/?page=2",
+      "previous": null,
+      "results": [
+        {
+          "id": "3Lw7OwK",
```
jgerigmeyer: @wlonk I imagine we should add a unique `slug` based on `product.name` for use in URLs? Probably also `old_slugs`...

metashare/api/views.py
stacyk: I wonder how often products would be hit compared to the other pages. My guess is people would spend the majority of their time on a project detail page. With that in mind, the negative of contacting GH each time the product list page is loaded seems like a smaller UX cost than having the user hit (hopefully hit) a button when they want to see any updates? 
It sounds like the consensus is to add a button at this point though.

jgerigmeyer: > I wonder how often products would be hit compared to the other pages. My guess is people would spend the majority of their time on a project detail page. With that in mind, the negative of contacting GH each time the product list page is loaded seems like a smaller UX cost than having the user hit (hopefully hit) a button when they want to see any updates?
> It sounds like the consensus is to add a button at this point though.

@stacyk I think either way we won't hit the endpoint every time the products-list page is loaded, but the question is whether we hit it on every page-refresh (typically once/session) or just on login (which may last through multiple sessions).

stacyk: @jgerigmeyer  To clarify, is this: 
> but the question is whether we hit it on every page-refresh (typically once/session)

not the same as:

> hit the endpoint every time the products-list page is loaded

jgerigmeyer: @stacyk Ah, well it could be or could not be, depending what you meant by "loaded". 😄 We certainly won't re-fetch during typical navigation through the application (like navigating to a product and then back to the products-list) -- which doesn't reload the browser. So it would only be on initial page-load or after a browser refresh/reload, vs only on login.

metashare/api/gh.py
```diff
@@ -0,0 +1,39 @@
...
+        url = url.path(url.path()[: -len(".git")])
+    if url.path().startswith("/repos"):
+        url = url.path(url.path()[len("/repos") :])
+    return str(url)
+
+
+def validate_gh_url(value):
+    if value != normalize_github_url(value):
+        raise ValidationError(
+            "%(value)s should be of the form 'https://www.github.com/:org/:repo'.",
```
jgerigmeyer: Can we also allow, for example, `http` or omitting `www` or whatnot?

metashare/api/views.py
```diff
@@ -1,19 +1,53 @@
...
+
+
+class ProductViewSet(viewsets.ModelViewSet):
+    permission_classes = (IsAuthenticated,)
+    serializer_class = ProductSerializer
+    pagination_class = ProductPaginator
+
+    def get_queryset(self):
+        repositories = self.request.user.repositories.values_list("url", flat=True)
+        return Product.objects.filter(repo_url__in=repositories)
```
jgerigmeyer: Is this comparing normalized versions of each url? Locally it's not working for me.

metashare/api/gh.py
wlonk: This is used by the admin to reject URLs without the proper form; we want to store them in a normalized form (and for me, that's `https`, `www`, and no `.git`) so we can query efficiently.

metashare/api/views.py
wlonk: This compares whatever is in the DB; as such, if you have older un-normalized product repo_urls in your DB, they won't match. I will double check that I'm normalizing the GH URLs, too.

metashare/api/gh.py
jgerigmeyer: Yeah, okay. Ideally we could make those adjustments automatically, but I guess that's not high priority for the admin.

metashare/api/views.py
jgerigmeyer: Yeah, it looks like (for me) the GH repos are still in the `https://api.github.com/repos/...` format.

metashare/api/gh.py
```diff
@@ -0,0 +1,47 @@
...
+    except (ObjectDoesNotExist, MultipleObjectsReturned):
+        return set()
+    gh = login(token=token)
+    repos = set(
+        [
+            normalize_github_url(repo.url)
+            for org in gh.organizations()
+            for repo in org.repositories()
+        ]
+        + [normalize_github_url(repo.url) for repo in gh.repositories()]
```
jgerigmeyer: @wlonk This is still only showing me user repos, not org repos. Might take a bit more digging?

mirisuzanne:
[I want to select from a list of products that I have contributor access to on github](https://trello.com/c/Xia3JFsW/11-i-want-to-select-from-a-list-of-products-that-i-have-contributor-access-to-on-github)

src/js/components/header/sfLogin.tsx
```diff
@@ -29,39 +28,52 @@ interface MenuOption {
...
-class Login extends React.Component<Props, { modalOpen: boolean }> {
-  public static defaultProps = {
-    id: 'login',
-    buttonClassName: 'slds-button_outline-brand',
-    buttonVariant: 'base',
-    disabled: false,
-    menuPosition: 'overflowBoundaryElement',
-    nubbinPosition: 'top right',
-  };
+const getMenuOpts = (): (MenuOption | Divider)[] => [
```
mirisuzanne: not sure where  or if I can see these options?

src/js/components/products/list.tsx
```diff
@@ -0,0 +1,197 @@
...
+            </div>
+          </>
+        );
+        break;
+      }
+      default: {
+        contents = (
+          <div className="slds-grid slds-wrap slds-grid_pull-padded-small">
+            {products.map(product => (
+              <ProductListItem product={product} key={product.id} />
```
mirisuzanne: I'm not clear  which github repos are being shown for what reasons. Maybe this is sample data, and I shouldn't worry about it?

src/sass/patterns/_containers.scss
```diff
@@ -0,0 +1,16 @@
+// Containers
+// ==========
+
+// Restrict fixed/absolute containers to half the viewport
+// ie. used on toast containers to unblock header links
```
mirisuzanne: How do I trigger a toast?

src/js/components/header/sfLogin.tsx
jgerigmeyer: Nope, you can't. :)

This file is hanging around but unused for the moment -- until we add SF login. (It's the same login dropdown as MD, functionally.)

src/js/components/products/list.tsx
jgerigmeyer: Products are shown if 1) they have been manually added to MetaShare by an admin, and 2) you are logged in and have contributor access on those repos.

This is sample data, with a number of SFDO and OddBird repos added.

src/sass/patterns/_containers.scss
jgerigmeyer: I think the simplest is to stop your server/container and the click the "Sync GitHub Repositories" button. (But again, functionally the toast is the same as on MetaDeploy.)

PR #48, opened 2019-07-02
[product-view] View product detail.

duggiemitchell:
See: https://trello.com/c/jG2pSNGW
cutedog


src/js/components/products/detail.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { useEffect, useState } from 'react';
+import { connect } from 'react-redux';
+// 
+
+const ProductDetail = () => (
+    <div>Hello world </div>
```
duggiemitchell: This is just a stand-in till I can get the component to display.

src/js/index.tsx
```diff
@@ -54,6 +55,11 @@ const App = () => (
               />
               <Route path={routePatterns.auth_error()} component={AuthError} />
               <PrivateRoute component={FourOhFour} />
+              <PrivateRoute
+                exact
+                path={routePatterns.product_detail()}
+                component={ProductDetail}
+              />
```
duggiemitchell: @jgerigmeyer am I missing something here?  Some of these patterns aren't in MD for me to follow, but, it's still showing the 404 page when routing to the newly created `ProductDetail` route.

src/js/components/products/listItem.tsx
```diff
@@ -7,7 +7,7 @@ import routes from '@/utils/routes';
 
 const ProductListItem = ({ product }: { product: Product }) => (
   <Link
-    to={routes.product_detail(product.slug)}
+    to={routes.product_detail(product.id)}
```
jgerigmeyer: I think we still want `slug` here.

src/js/index.tsx
jgerigmeyer: @duggiemitchell This route needs to come _before_ the 404 catch-all route (probably should come right after `product_list`). Because the 404 has no `path` defined, it matches any and all urls that get to that point.

src/js/utils/routes.ts
```diff
@@ -12,7 +12,7 @@ export const routePatterns = {
   login: () => '/login',
   auth_error: () => '/accounts/*',
   product_list: () => '/products',
-  product_detail: () => '/products/:productSlug',
+  product_detail: () => '/products/:productId',
```
jgerigmeyer: Stick with `productSlug` here.

src/js/index.tsx
duggiemitchell: OK thank you, that seemed to do it; I wasn't sure whether the order mattered.

src/js/components/products/detail.tsx
```diff
@@ -17,21 +18,19 @@ import {
   // selectProductNotFound,
 } from '@/store/products/selectors';
 
-interface Props {
+type Props = {
   product: Product;
   productSlug: string;
-}
+} & RouteComponentProps;
```
duggiemitchell: @jgerigmeyer  I was thinking this needs to be `RouteChildrenProps` that are defined in  `@types/react-router-dom` but for some reason, but it's saying `@types/react-router-dom"' has no exported member 'RouteChildrenProps'`.  🤷🏽‍♀️

src/js/components/products/detail.tsx
```diff
@@ -40,25 +39,27 @@ const ProductDetail = ({ product, productSlug }: Props) => {
...
-                [{product.name}]
+                [{i18n.t(product.name)}]
               </h2>
               <p
                 className="markdown slds-p-bottom_small"
                 dangerouslySetInnerHTML={{ __html: product.description }}
               />
-              <a href={product.repo_url}>Link to Github Repo</a>
+              <a href={product.repo_url}>
+                <Trans i18nKey="githubRepoLink">Link to Github Repo</Trans>
```
duggiemitchell: There's some "missing translations" errors.  Is there an additional step to add translations?  Do we commit the changes to `/locales_dev/en/translation.json`

duggiemitchell: Nevermind, about translations, looks like that is covered [here in the read.me](https://github.com/SFDO-Tooling/MetaShare/blob/master/CONTRIBUTING.rst#internationalization)

src/js/components/products/detail.tsx
```diff
@@ -0,0 +1,78 @@
...
+} from '@/store/products/selectors';
+import routes from '@/utils/routes';
+
+type Props = {
+  product?: Product | null;
+  productSlug?: string;
+} & RouteComponentProps;
+
+const ProductDetail = ({ product, productSlug }: Props) => (
+  <DocumentTitle title={`${i18n.t(productSlug)} | ${i18n.t('MetaShare')}`}>
```
jgerigmeyer: No need to run translation on values that come from the database -- if they need to be translated, that'll happen on the DB side. So we only wrap hard-coded front-end strings in translation flags.

src/js/components/products/detail.tsx
```diff
@@ -0,0 +1,78 @@
...
+            </h2>
+            {/* This description is pre-cleaned by the API */}
+            {product.description && (
+              <p
+                className="markdown slds-p-bottom_small"
+                dangerouslySetInnerHTML={{ __html: product.description }}
+              />
+            )}
+            <a href={product.repo_url}>
+              <Trans i18nKey="githubRepoLink">Link to Github Repo</Trans>
```
jgerigmeyer: The `<Trans>` block is useful in two cases:

1. Really really long strings.
2. Translations that contain other inline elements (e.g. <a>, <strong>, etc).

Since this has neither, it's easier just to use `i18n.t(...)`.

```suggestion
              {i18n.t('Link to Github Repo')}
```

jgerigmeyer:
[I want to view info for an existing product (see new product for details).](https://trello.com/c/jG2pSNGW/12-i-want-to-view-info-for-an-existing-product-see-new-product-for-details)

src/js/components/products/detail.tsx
```diff
@@ -0,0 +1,124 @@
...
+              <h2 className="slds-text-heading_medium">{product.name}</h2>
+            )}
+            {/* This description is pre-cleaned by the API */}
+            {product.description && (
+              <p
+                className="markdown"
+                dangerouslySetInnerHTML={{ __html: product.description }}
+              />
+            )}
+            <a href={product.repo_url}>{i18n.t('Link to GitHub Repo')}</a>
```
mirisuzanne: Do we need to say "link to"?

src/js/components/products/detail.tsx
```diff
@@ -0,0 +1,124 @@
...
+            className="slds-col
+              slds-size_1-of-1
+              slds-medium-size_1-of-3
+              slds-text-longform"
+          >
+            {!productDescriptionHasTitle && (
+              <h2 className="slds-text-heading_medium">{product.name}</h2>
+            )}
+            {/* This description is pre-cleaned by the API */}
+            {product.description && (
```
mirisuzanne: I'm seeing Lorem Ipsum, is that right for now?

src/sass/components/_breadcrumbs.scss
```diff
@@ -0,0 +1,3 @@
+.slds-breadcrumb__item {
```
mirisuzanne: Careful about overriding slds classes – better to add a new class?

src/js/components/products/detail.tsx
duggiemitchell: @mirisuzanne I think this is fine, sample data from backend?

src/sass/components/_breadcrumbs.scss
jgerigmeyer: In MD we're doing this same override, but only within a custom class: https://github.com/SFDO-Tooling/MetaDeploy/blob/master/src/sass/components/_header.scss#L16-L32

Maybe we should take a similar approach here, nesting this within `.ms-breadcrumb` or something like that?

src/js/components/products/detail.tsx
stacyk: ha, it is quite redundant for screen readers if so, and sighted users will see a color change/underline so I think it is safe to remove "link to"

metashare/api/tests/models.py
```diff
@@ -1,5 +1,28 @@
+from unittest.mock import patch
+
 import pytest
 
+from ..models import Product, ProductSlug, user_logged_in_handler
+
+
+@pytest.mark.django_db
+class TestProductSlug:
```
duggiemitchell: do we need to test for productslug?

metashare/api/views.py
```diff
@@ -1,19 +1,58 @@
...
+        user = self.get_object()
+        repos = gh.get_all_org_repos(user)
+        GitHubRepository.objects.filter(user=user).delete()
+        GitHubRepository.objects.bulk_create(
+            [GitHubRepository(user=user, url=repo) for repo in repos]
+        )
+        serializer = FullUserSerializer(user)
+        # TODO: this _should_ return the new list of products for the user, but that
+        # feels wrong for this endpoint. Merits thought.
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
duggiemitchell: changing status code to 204

src/js/components/products/detail.tsx
jgerigmeyer: @stacyk @mirisuzanne Thoughts?

Screenshot 2019-07-09 14 54 39

stacyk: should the icon be the link color as well (and also inherit hover/focus color)?

jgerigmeyer: @stacyk Good call.

Screenshot 2019-07-09 15 46 28

PR #49, opened 2019-07-02
[persist-database-locally] Keep the database data locally

wlonk:
This should make a gitignored postgres directory, and save your db data in it, between calls to down, build, etc.

You will want to run:

```
make down
make build
make up
```

to get these changes in locally, I think.

red wolf

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ yarn-error.log
 staticfiles/
 dump.rdb
 /env
+/postgres
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,6 +29,8 @@ services:
     image: postgres:10.6
     environment:
       - POSTGRES_USER=metashare
+    volumes:
+      - ./postgres:/var/lib/postgresql/data
 
   # Runs the queue process:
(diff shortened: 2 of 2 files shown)


jgerigmeyer:
> @wlonk still getting the `DoesNotExist at /accounts/github/login/` when trying to login on this branch after these commands:
> 
> make down
> make build
> make up

@duggiemitchell Just to be sure, you'll have to run the `populate_...` script once. But then if you can login successfully, run `make down; make build; make up` again and see if it still works?

duggiemitchell:
> > @wlonk still getting the `DoesNotExist at /accounts/github/login/` when trying to login on this branch after these commands:
> > make down
> > make build
> > make up
> 
> @duggiemitchell Just to be sure, you'll have to run the `populate_...` script once. But then if you can login successfully, run `make down; make build; make up` again and see if it still works?

OK, I think this is good.  The database seems to survive a `make down` though it does still have issues with my account, likely related to having an additional email; though I don't believe this branch has the code that was added as a workaround for that, correct @jgerigmeyer @wlonk ?

wlonk:
@duggiemitchell Correct; this is just for DB persistence. Working on diagnosing the login issues with @stacyk separately.

PR #51, opened 2019-07-03
[fix-webpack-error-falldown] Retry failed web containers

wlonk:
beelzebub as a cat

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,11 @@ services:
       - ~/.gitconfig:/root/.gitconfig
     ports:
       - "0.0.0.0:8080:8080"
+    restart_policy:
+      condition: on-failure
+      delay: 5s
+      max_attempts: 3
+      window: 120s
 
   # Runs the database process:
   postgres:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,7 @@ services:
       - postgres
       - redis
     links:
-      - "postgres:db"
+      - 'postgres:db'
       - redis
     env_file: .env
     environment:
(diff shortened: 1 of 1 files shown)


wlonk:
> (Even if it did, I assume it would fail again unless the error was quickly fixed in the meantime.)

Yeah; in our old style, you just re-run `yarn` if it has a fatal error, so I don't know what we might expect this to do differently.

`sleep infinity` is fine if you want the container to be present but not doing anything, as VSCode does, but we have to want it to do something for dev. I guess I'm really not sure what your ideal might be here.

jgerigmeyer:
> I guess I'm really not sure what your ideal might be here.

Well in VS Code when the container falls down everything breaks -- because you're working in a window running _in the container_, you can no longer make any changes to files, etc. So you can't even fix the error that caused the falldown. So currently it's pretty bad.

I guess my ideal would be that the container keeps running, even if it's not doing anything. That would allow fixing the error and then running a command or script (within the container shell) to restart the server/watcher?

wlonk:
Oh, I see. It's a lot like VSCode wants a VM, not a container, so the VM can continue, and can also run other commands that may start, stop, or crash as they do.

Why does VSCode not just run an instance of the container with `sleep infinite` as the command, and separately, you run `docker-compose up` or whatever to get the container that's doing the yarn building and serving? I ask "why not" as an honest question—I presume there's a reason, but I don't know it yet.

jgerigmeyer:
@wlonk Yeah, I can't honestly say I understand the intricacies of how it works -- or even how Docker works -- so I'm just going based on what I've gathered from the (relatively good) VSC/Docker docs, and how it seems to work in practice. ¯\_(ツ)_/¯

wlonk:
On Monday, let's do a screenshare and you can walk me through what VSCode does and how the workflow looks, so I can better understand.

jgerigmeyer:
@wlonk I think this is made somewhat moot by the changes merged in #72. Is there still a reason we want this PR?

wlonk:
Affirmatively no.

PR #69, opened 2019-07-08
[project-create] Create a project

wlonk:
See https://trello.com/c/eWmkxO8p/18-i-want-to-create-a-new-project-for-a-specific-product-giving-it-a-name-and-a-general-description-of-the-changes-i-plan-to-make#

sea lion


metashare/api/models.py
```diff
@@ -129,6 +129,22 @@ def __str__(self):
...
+class ProjectSlug(AbstractSlug):
+    parent = models.ForeignKey(
+        "Project", on_delete=models.PROTECT, related_name="slugs"
+    )
+
+
+class Project(mixins.HashIdMixin, mixins.TimestampsMixin, SlugMixin, models.Model):
+    name = models.CharField(max_length=50, unique=True)
+    description = MarkdownField(blank=True, property_suffix="_markdown")
+    pr_url = models.URLField(unique=True)
```
jgerigmeyer: @wlonk Projects will always have a `branch_url` (not sure if we need to store that or not). I'm not clear on whether we're also creating a PR immediately when the Project is created (and then just marking it "ready for review" somehow on the "capture changes" action), or waiting until "capture changes" to create the PR at all.

metashare/api/models.py
```diff
@@ -138,7 +138,7 @@ class ProjectSlug(AbstractSlug):
 class Project(mixins.HashIdMixin, mixins.TimestampsMixin, SlugMixin, models.Model):
     name = models.CharField(max_length=50, unique=True)
     description = MarkdownField(blank=True, property_suffix="_markdown")
-    pr_url = models.URLField(unique=True)
+    pr_url = models.URLField(unique=True, null=True)
     commit_message = MarkdownField(blank=True, property_suffix="_markdown")
     release_notes = MarkdownField(blank=True, property_suffix="_markdown")
```
duggiemitchell: @wlonk We are asking for just the `name` and `description` field for initial project creation, so I think the other fields can also be null.

metashare/api/views.py
```diff
@@ -42,9 +42,16 @@ class ProductViewSet(viewsets.ModelViewSet):
...
     def get_queryset(self):
         repositories = self.request.user.repositories.values_list("url", flat=True)
         return Product.objects.filter(repo_url__in=repositories)
+
+
+class ProjectViewSet(viewsets.ModelViewSet):
+    permission_classes = (IsAuthenticated,)
+    serializer_class = ProjectSerializer
+    pagination_class = CustomPaginator
+    queryset = Project.objects.all()
```
jgerigmeyer: @wlonk Are these ordered by most-recently-created-first? I think that's the desired functionality, at least for now.

src/js/components/projects/projectForm.tsx
```diff
@@ -0,0 +1,94 @@
+/* eslint-disable @typescript-eslint/camelcase */
```
jgerigmeyer: I'm not sure why this is needed? Or if it is, it'd be nice to only use it for the relevant code section.

src/js/components/projects/projectForm.tsx
```diff
@@ -0,0 +1,94 @@
...
+// @todo can i use connect if I dont need the state?  just to create an action.
+// const select = (appState: AppState) => ({
+//   state: appState,
+// });
+
+const actions = {
+  doCreateProject: createProject,
+};
+const WrappedProjectForm = connect(
+  () => {},
```
jgerigmeyer: Pass `null` here to skip state subscriptions: https://react-redux.js.org/api/connect#mapstatetoprops-state-ownprops-object

metashare/api/fields.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.fields import CharField
+
+
+class MarkdownField(CharField):
+    def to_representation(self, value):
+        # This presumes that you are bound to a serializer and have an
+        # instance on that serializer, neither of which I like, but it
+        # does avoid duplicating the logic in the model MarkdownField.
+        return getattr(self.parent.instance, self.field_name + "_markdown")
```
davisagli: Should this be in sfdo-template-helpers?

metashare/api/models.py
davisagli: Wait until changes are captured, I think.

Should we just store a PR id and synthesize the rest of the URL? If we move a repository to a different github organization, I want to have to update the repository URL in exactly one place in MetaShare.

metashare/api/models.py
```diff
@@ -132,6 +132,22 @@ def __str__(self):
         return self.url
 
 
+class ProjectSlug(AbstractSlug):
+    parent = models.ForeignKey(
+        "Project", on_delete=models.PROTECT, related_name="slugs"
+    )
+
+
+class Project(mixins.HashIdMixin, mixins.TimestampsMixin, SlugMixin, models.Model):
```
jgerigmeyer: @wlonk I don't see anywhere that a `project` is owned by a `product` -- I think we need a relation here?

src/js/components/projects/projectForm.tsx
duggiemitchell: It's there since the api is requiring params in camelCase to create a project (I'm guessing temporarily) such as the `commit_message` and `release_notes` as seen here: https://github.com/SFDO-Tooling/MetaShare/pull/69/commits/9d37af049d0e7857e1efa7fc60707824e0a5c7dd#diff-177bdc19a36cd6b157d923a44b273f69R32

src/js/components/projects/projectForm.tsx
duggiemitchell: :thumb

metashare/api/fields.py
wlonk: I'd like to work out the kinks in it a bit here before moving it to helpers.

metashare/api/models.py
wlonk: Yeah, not sure how I missed that besides general high levels of stress and ball-dropping last week :(

metashare/api/models.py
wlonk: @duggiemitchell `blank=True` should account for that; I'd rather that they be empty strings than null, unless null means something distinct from an empty string.

metashare/api/views.py
wlonk: Implicitly, yes. They should get an explicit ordering, though.

metashare/api/models.py
jgerigmeyer: @wlonk I don't think the front end should need to know to add in keys with empty values to the `POST` data. We're using a form, so IMO the back end should be able to handle a `POST` that only includes the data that the user or front end needs to provide, e.g.:

```
{
  'name': 'My Project',
  'description': '',
}
```

wlonk: Right; so missing fields + defaults should handle that.

wlonk: (That is, defaults on the backend.)

src/js/components/projects/projectForm.tsx
```diff
@@ -26,11 +28,13 @@ const ProjectForm = ({ productName, doPostObject }: Props) => {
...
     }
-    // check if project (name) exists?
     doPostObject({
       objectType: OBJECT_TYPES.PROJECT,
       content: { name, description, commit_message: '', release_notes: '' },
-    });
+    }).finally(() => (
+      // display errors i.e. if project (name) exists?
+      <Redirect to={routes.project_detail(productSlug, productName)} /> // @todo, this isn't redirecting...
+    ));
```
duggiemitchell: @jgerigmeyer a couple things about these couple lines:  
1. Any idea why this redirect is not working? 
2.  The api returns error msgs that are stored in redux.  Would using a selector be the approach for displaying those errors? (i.e. if the `name` already exists for a project).
3.  Would we need need to anticipate long descriptions or project names? I don't see a char limit on those. 
4.  Are there restricted chars for project names?

wlonk:
@jgerigmeyer I think I addressed all the points you raised.

metashare/api/models.py
wlonk: Eugh, I added it and made a migration and everything, what happened. I'll fix it tonight.

src/js/components/projects/projectForm.tsx
jgerigmeyer: 1. `<Redirect>` is a _declarative_ way to handle routing redirects, but that will only work within a `render` block (that is, rendered markup). Instead here you'll want to use `withRouter()` and then `history.push()` to navigate. See e.g. https://github.com/SFDO-Tooling/MetaDeploy/blob/master/src/js/components/plans/ctaButton.js#L215-L228

Related, I think we only want to do that redirect on success (`.then()`), not `.finally()`.

2. Hm, yeah. I don't think we want to store form-specific error messages globally, no. My hunch is that we want to do two things: 1) Make sure the API is returning helpful human-readable error messages in the `details` or `non_field_errors` keys, and 2) Expose field-specific error messages to the component, storing them to display in local component state (not the global redux store). That would also allow us to get rid of the custom `hasName` logic so we can just rely on server-side validation.

I'd be happy to tackle this problem, if you'd like. Or we can talk more about what I mean.

3. Yeah, good questions. @wlonk What do you think about restrictions on length or specific chars?
4. ☝️

src/js/components/projects/projectForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+      {projectCreateActive && (
+        <>
+          <h1>Create a Project for {productName}</h1>
+
+          <form onSubmit={handleSubmit}>
+            <Input
+              id="project-name"
+              label="Project Name"
+              required
+              errorText={hasName ? null : 'Project name is required.'}
```
jgerigmeyer: These strings will need to be translated (also `label` above and `label` below).

src/js/components/projects/projectForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+          </form>
+        </>
+      )}
+
+      <Button
+        label={i18n.t('Create a Project')}
+        className="slds-size_full slds-p-vertical_xx-small"
+        variant="brand"
+        onClick={formControl}
+      />
```
jgerigmeyer: Looks like we're missing the close/cancel button?

src/js/components/projects/projectForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const [description, setDescription] = useState('');
+  const [name, setName] = useState('');
+  const [projectCreateActive, setprojectCreateActive] = useState(false);
+  const [hasName, setHasName] = useState(true);
+
+  const handleSubmit = () => {
+    if (name === '' && hasName) {
+      setHasName(false);
+      return;
+    }
```
jgerigmeyer: I'm not clear on this `setHasName` logic throughout. It seems odd to me that the first time I try to submit without a name I get a warning, but then on subsequent times it just goes through and tries anyway. I'd say let's either prevent consistently, or allow the server-side validation to do its thing.

src/js/store/actions.ts
```diff
@@ -37,21 +37,38 @@ interface FetchObjectFailed {
...
+  type: 'POST_OBJECT_STARTED';
+}
+
+interface PostObjectSucceeded {
+  type: 'POST_OBJECT_SUCCEEDED';
+}
+
+interface PostObjectfailed {
+  type: 'POST_OBJECT_FAILED';
+}
```
jgerigmeyer: These types are missing payloads.

src/js/utils/routes.ts
```diff
@@ -5,6 +5,8 @@ const routes = {
   login: () => '/login',
   product_list: () => '/products',
   product_detail: (productSlug: string) => `/products/${productSlug}`,
+  project_detail: (productSlug: string, projectName: string) =>
+    `/products/${productSlug}/${projectName}`, // using projectname for now
```
jgerigmeyer: Ideally we want `projectSlug` here.

src/js/store/projects/reducer.ts
```diff
@@ -0,0 +1,4 @@
+export interface Project {
+  name: string;
+  description: string;
+}
```
jgerigmeyer: For each product, we’ll want a list of `projects` and also a `next` url (maybe also a `notFound` list, but that can wait for now). We _could_ store that data on individual products, but I wonder if it might be cleaner to have a separate `projects` top-level state object, but then have it separated by product (probably `id`?)…

```
{
  products: ...,
  projects: {
    [product.id]: {
      projects: [...],
      next: null,
    },
    ...
  }
}
```

metashare/api/models.py
wlonk: @jgerigmeyer fixed.

src/js/components/projects/projectForm.tsx
jgerigmeyer: Some further clarification on points 3-4: Project names are already limited to 50 chars. Descriptions might get long, but that's okay -- we can truncate them for display purposes where we need to (e.g. the list view). And we don't need to worry about restrictions -- we use the `name` to create a `slug` (which is already a munged version), and we can use that for the GH branch name.

src/js/components/products/detail.tsx
```diff
@@ -51,12 +58,39 @@ const ProductDetail = ({ product, productSlug, doFetchObject }: Props) => {
     product.description &&
     (product.description.startsWith('<h1>') ||
       product.description.startsWith('<h2>'));
+
+  // some mock data for now
```
jgerigmeyer: The API is in place to get real data; you should be able to use `fetchObjects` with `objectType: 'project'` instead of this mock data.

metashare/api/filters.py
```diff
@@ -9,3 +9,9 @@ class ProductFilter(filters.FilterSet):
     class Meta:
         model = Product
         fields = ("slug",)
+
+
+class ProjectFilter(filters.FilterSet):
+    class Meta:
+        model = Project
+        fields = ("product",)
```
jgerigmeyer: If the `?product=` query string is not a valid product, this produces an unexpected `400` error:

```
{"product":["Select a valid choice. That choice is not one of the available choices."]}
```

I guess that shouldn't really happen in practice, but it does strike me as surprising.

src/js/components/products/detail.tsx
jgerigmeyer: @duggiemitchell Also remember that you'll want to filter each projects query by product, e.g.:

```js
fetchObjects({
  objectType: 'project',
  filters: {
    product: product.id,
  },
  reset: true,
})
```

src/js/components/projects/listItem.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Product } from '@/store/products/reducer';
+import { Project } from '@/store/projects/reducer';
+import routes from '@/utils/routes';
+
+interface Props {
+  product: Product;
+  project: Project;
+}
+
+const ProjectListItem = ({ product, project }: Props) => {
```
stacyk: @duggiemitchell Can we make this into a list? I don't know where hte <ul> would go here since this is a repeated block and I only want one ul wrapper element, but I'm looking at something like: 
```
<ul className="slds-has-dividers_bottom-space">
    <li className="slds-item project-list-item">
      {/* @todo capture slug from actual project */}
      <h2 className="slds-text-heading_medium">
        <Link to={routes.project_detail(product.slug, slug)}>
          {name}
        </Link>
      </h2>
      {description && (
        <p
          className="markdown project-list-description"
          dangerouslySetInnerHTML={{ __html: description }}
        />
      )}
      </li>
  </ul> 
```

docker-compose.yml
```diff
@@ -23,6 +23,9 @@ services:
       - ~/.gitconfig:/root/.gitconfig
     ports:
       - '0.0.0.0:8080:8080'
+      - '0.0.0.0:33284:33284'  # Spells "DEBUG" in phone-number
```
jgerigmeyer: What is this for?

metashare/api/serializers.py
```diff
@@ -29,3 +31,36 @@ class Meta:
...
+        return f"{obj.product.repo_url}/tree/{obj.branch_name}"
+
+    def create(self, validated_data):
+        instance = super().create(validated_data)
+        gh.create_branch(
+            self.context["request"].user,
+            instance.product.repo_url,
+            instance.branch_name,
+        )
+        return instance
```
jgerigmeyer: @wlonk Given that we now want to wait on branch creation until the project has a task with a scratch org... should we comment out this block?

src/js/components/products/detail.tsx
```diff
@@ -32,7 +45,14 @@ const ProductDetail = ({ product, productSlug, doFetchObject }: Props) => {
         filters: { slug: productSlug },
       });
     }
-  }, [product, productSlug, doFetchObject]);
+
+    if (product) {
```
jgerigmeyer: I don't think we want to do this every time the product changes... so we'll probably want to check and see if we've already fetched the projects for this product? E.g. once you have the `projects` state in this component, something like:

```js
if (product && !projects[product.id]) { ... }
```

metashare/api/serializers.py
wlonk: Yeah, we should.

docker-compose.yml
wlonk: To get a Python debugger in a Docker process, the best way is to use a remote debugger, that talks over an open port. I guess "Spells DEBUG in phone-number" doesn't adequately explain that 😆

src/js/components/products/detail.tsx
```diff
@@ -84,15 +114,35 @@ const ProductDetail = ({ product, productSlug, doFetchObject }: Props) => {
...
+                    <ul className="slds-has-dividers_bottom">
+                      {projects.projects.map(project => (
+                        <ProjectListItem
+                          key={project.id}
+                          project={project}
+                          product={product}
+                        />
+                      ))}
+                    </ul>
+                  </>
```
jgerigmeyer: I think we also want prev/next pagination links here? Could happen in a new branch, since this technically works without it.

src/js/components/products/detail.tsx
```diff
@@ -84,15 +114,35 @@ const ProductDetail = ({ product, productSlug, doFetchObject }: Props) => {
...
+                <ProjectForm
+                  product={product}
+                  startOpen={!projects.projects.length}
+                />
+                {Boolean(projects.projects.length) && (
+                  <>
+                    <h2 className="slds-text-heading_medium">
+                      {i18n.t('Projects for')} {product.name}
+                    </h2>
+                    <ul className="slds-has-dividers_bottom">
```
jgerigmeyer: Do we want to hide this divider for the last item in the list?

metashare/api/fields.py
```diff
@@ -0,0 +1,44 @@
+import bleach
+from markdown import markdown
+from rest_framework.fields import CharField
+
+# These two constants are duplicated from sfdo-template-helpers' MarkdownField.
+# Not the best thing, but the clearest way to render markdown out of a serializer.
```
jgerigmeyer: Yeah, this duplication is unfortunate. Is there a way we can clean it up?

src/js/components/products/detail.tsx
stacyk: no, I mean, it will make more sense when pagination is in so I'd keep it for now

stacyk:
> lgtm - the one issue I noticed doesn't seem to be new in this branch: we have two `h1` elements on each page. Generally only the page title should be `h1`, not the site title.

I only see one h1 (page title). The "metaSHARE" is in a div for me. 

Oh, I see. If a person uses an H1 in the description (like on the MetaShare product) then what should we do?

stacyk:
> lgtm - the one issue I noticed doesn't seem to be new in this branch: we have two `h1` elements on each page. Generally only the page title should be `h1`, not the site title.

@mirisuzanne Is this what you are referring to or is there a different h1 I am not seeing?

2-h1s

mirisuzanne:
@stacyk sure, that too. But I'm seeing an `h1` on `metaSHARE`:

metaSHARE

stacyk:
> @stacyk sure, that too. But I'm seeing an `h1` on `metaSHARE`:
> 
> metaSHARE

hm, interesting. Is that only on the homepage for you? I get nested divs with only an <a>, no <h1> 
nested-divs-no-h1

mirisuzanne:
@stacyk nope, I see it on every page

jgerigmeyer:
@mirisuzanne @stacyk Regarding the `<h1>`... that's out of our control at the moment, added by SLDS-React to every PageHeader title (https://github.com/salesforce/design-system-react/blob/master/components/page-header/private/title.jsx#L57-L71). We could try using a different component or submitting a PR.

jgerigmeyer:
Closing in favor of #94.

PR #71, opened 2019-07-10
[dry-actions] 🧶 Abstract generic fetchObject(s) actions.

jgerigmeyer:
The goal is to create reusable actions for most object types. We could probably add a generic `createObject` action too, and `deleteObject` if we need it.

drying dog

### DIFF
--- a/src/js/store/products/reducer.ts
+++ b/src/js/store/products/reducer.ts
@@ -1,5 +1,7 @@
+import { ObjectsAction } from '@/store/actions';
 import { ProductsAction } from '@/store/products/actions';
 import { LogoutAction } from '@/store/user/actions';
+import { OBJECT_TYPES } from '@/utils/constants';
 
 export interface Product {
   id: string;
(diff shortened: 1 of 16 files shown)


jgerigmeyer:
> I think this looks good and clean. Idk how I feel about the name being `object(s)` but if generic is what we are going for, that is it.

Yeah, I'm open to other suggestions. I'd like these actions to be used for various things (e.g. products, projects, tasks, etc) -- and "object" was the best I could come up with. 🤷‍♂

> One question I have is what does the new `reset` boolean flag do?

Again, I'm open to other suggestions for structure or names. It flags the difference between "I'm fetching a list of things, and the results should override whatever we already know about" vs "I'm fetching a list of things (e.g. page 2), and the results should be added to whatever we already know about".

src/js/store/products/reducer.ts
```diff
@@ -24,39 +26,51 @@ const defaultState = {
 
 const reducer = (
   products: ProductsState = defaultState,
-  action: ProductsAction | LogoutAction,
+  action: ProductsAction | ObjectsAction | LogoutAction,
```
wlonk: Still using `ProductsAction` here?

jgerigmeyer: There is still one custom `ProductsAction`, for syncing the repos.

davisagli:
Nice!

PR #83, opened 2019-07-12
[helper-script] Use a helper script instead of a Makefile

wlonk:
The motivation being that it's a lot easier, this way, to pass additional arguments into the command.

hummingbird


davisagli:
I like the idea and follow the logic of the name, but how about `ship` as a command that's shorter to type.

wlonk:
@davisagli Amusingly, perhaps, that's what I had called it initially, but the more I tried it out with that name, the more it felt like it was colliding with the sense of "shipping software".

I can still use `ship`, if you'd rather.

davisagli:
@wlonk `dinghy` then. But I'm fine sitting with this for now.

PR #87, opened 2019-07-16
[add-welcome-mat] Add welcome mat

jgerigmeyer:
https://trello.com/c/YG9GLBf4

great dane


jgerigmeyer:
[I want to log in with a github account](https://trello.com/c/YG9GLBf4/10-i-want-to-log-in-with-a-github-account)

stacyk:
I think this one is ready to review. Assigning @mirisuzanne and @duggiemitchell

src/js/components/user/login.tsx
```diff
@@ -53,10 +53,11 @@ const Login = ({ user }: { user: User | null }) =>
   user ? (
     <Redirect to={routes.home()} />
   ) : (
-    <div className="slds-welcome-mat slds-welcome-mat_info-only">
-      <div className="slds-welcome-mat__content slds-grid">
+    <div className="slds-welcome-mat slds-welcome-mat_info-only welcome-container">
```
duggiemitchell: Any reason for not using the react [`<WelcomeMat/>` component](https://react.lightningdesignsystem.com/components/welcome-mat/) instead?

stacyk:
> This looks good to me. Any way of getting the welcome mat image to be be the full height? maybe adding an additional class on the same div as the grid column?

It is full height in my browser so I'm guessing whichever you are checking in that is ignoring some of my overrides. Can you send a screenshot please? 

As far as the modal component, we have nothing to show under the modal. We thought it might be a bit odd to have the ability to close a modal to only have to re-open it right away to login. ¯\_(ツ)_/¯

duggiemitchell:
That makes sense.  Here's how it looks for me (on Chrome):
Screenshot from 2019-07-18 16-04-52

src/js/components/user/login.tsx
```diff
@@ -50,23 +53,79 @@ const Login = ({ user }: { user: User | null }) =>
...
+            slds-size_1-of-1
+            slds-medium-size_1-of-2
+            slds-welcome-mat__tiles_info-only
+            slds-grid
+            slds-grid_vertical
+            slds-grid_align-center
+            slds-p-left_xx-large
+            slds-p-right_xx-large"
+        >
+          <WelcomeMatTile
```
mirisuzanne: Clearly these tiles need content - but this first "welcome" tile feels likely redundant with the "welcome" info we already have on the left?

stacyk: Agreed. Everything here is placeholder as of now including the topics for the blocks of text. I do think if there isn't an obvious set of instructions, we can answer the who, what, when/where, why and mostly how types of questions someone might have. If this isn't helpful, then we would need to re-think this side of the mat. I've never written copy for a client though so if that is on us, I think I'd assign that to someone on our team that does have a writing background. Or maybe the SF team wants to fill these blocks in? I'm not sure where we stand on that.

jgerigmeyer: Those are good questions. I'd say given this is an in-development app, let's merge this and then discuss copy (and who should write it) with Jason and David?

jgerigmeyer:
@duggiemitchell Can you confirm that the image-height issue is fixed?

duggiemitchell:
@stacyk It's still not showing up full height for me in chrome, but :+1: in firefox.
Screenshot from 2019-07-24 12-03-20

src/js/components/user/login.tsx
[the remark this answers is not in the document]
mirisuzanne: The react component requires the welcome mat to be in a modal - and we wanted it full-page.

PR #94, opened 2019-07-23
[project-pagination] Paginate project list 

duggiemitchell:
See Trello: https://trello.com/c/XTJkvpAe branched from: https://trello.com/c/rXpr25Wc

cute-baby-animals-32


duggiemitchell:
@jgerigmeyer I figured the best way to write this without having enough test projects is by copying the existing patterns for pagination in the product detail component and writing tests to make sure it works right. Is that the right approach?  I'd like some feedback on what you think is wrong with this test setup and if this is even the right approach to getting this done.  Also, do you think it would be better to separate the function for scroll tracking to a utility helper for better reuse?

jgerigmeyer:
@duggiemitchell Looking at the mockup (https://xd.adobe.com/view/2dc1958f-96ed-4f7a-6140-87fe98ae0813-5bf0/), the pagination here is not scroll-based -- so it won't follow the same pattern as products. It's explicit `prev` and `next` links. The action will be similar, but we'll probably want to do something where we track which page we're showing. But that means just storing the `next` url in the store won't be sufficient.

This will take a bit more thought. If you want to make a quick proposal for how to architect this that'd be fine -- otherwise let's put it on the back-burner for now.

duggiemitchell:
@jgerigmeyer OK I didn't see the updated mockup with pagination.  I can give it some thought to a proposal for this, thanks!

duggiemitchell:
@jgerigmeyer this new commit has the changes for pagination with `previous` `next` links. 
> but we'll probably want to do something where we track which page we're showing. 

I think this could be accomplished through adding a page number to the url.  The project results return a count, and page max is 25.  So we can do some math to get the current page.  Though if a project is created and we are on page 3, a user would have to go to page 1 in order to see it in the list.

> And I guess we'll also need to track whether the page being requested has already been fetched or not.

My first guess is to use the sessionstorage to 'cache' page results, so when someone clicks next/previous, it loads from the storage. But how would we track whether a new item is in the list not saved in storage ?

stacyk:
I don't know how this impacts @duggiemitchell and @jgerigmeyer's planning on this, but the last place we left this I think is following @mirisuzanne's suggestion here: https://trello.com/c/rXpr25Wc/20-i-want-to-select-a-project-from-a-list-of-projects-for-a-specific-product-that-i-have-access-to#comment-5d13e24893cbf314e3d761dd using an even more simplified "load more" instead of previous and next links. I didn't update the prototype but I should have updated the description. That said, "load more" is harder to find an older project if you know the project you are looking for is far down in the list. I like the idea of having an option to recall a "page 3" type of url even if it changes based on how many new projects are being added. That said, ideally I'd like the numeric pagination BUT since we are looking to streamline things at these MVP stages, I would still vote for the prev / next if it does update the URL. Otherwise, load more is fine, and I would up the limit to something much higher since it doesn't really effect performance when you load 50 text links, as opposed to 25, does it?

duggiemitchell:
This should be functionally complete.  @stacyk I assigned you to give the styling a once/over approval.

stacyk:
[I want project pagination.](https://trello.com/c/XTJkvpAe/57-i-want-project-pagination)

jgerigmeyer:
@duggiemitchell Nice work! This looks (and works) great 👏

jest.config.js
```diff
@@ -25,20 +25,18 @@ module.exports = {
...
   coverageDirectory: 'coverage',
 
   // An array of regexp pattern strings used to skip coverage collection
-  // coveragePathIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  coveragePathIgnorePatterns: [
+    'src/js/index.tsx',
+    'src/js/sentry.ts',
+    'src/js/i18n.ts',
```
wlonk: I'm a little leery of ignoring `src/js/index.tsx` and `src/js/i18n.ts` as I expect them to have meaningful code in them; is the only way to exercise them to run the whole project or something?

src/js/components/products/detail.tsx
```diff
@@ -34,6 +56,17 @@ const ProductDetail = ({ product, productSlug, doFetchObject }: Props) => {
...
+  useEffect(() => {
+    if (product && (!projects || !projects.fetched)) {
+      // Fetch projects from API
+      doFetchObjects({
+        objectType: OBJECT_TYPES.PROJECT,
+        filters: { product: product.id },
+        reset: true,
+      });
+    }
+  }, [product, projects, doFetchObjects]);
```
wlonk: I think I don't adequately understand `useEffect`; what are the values in the trailing array, and how does this set of them make it behave as we expect?

src/js/components/products/detail.tsx
```diff
@@ -84,15 +140,53 @@ const ProductDetail = ({ product, productSlug, doFetchObject }: Props) => {
...
+            {!projects || !projects.fetched ? (
+              // Fetching projects from API
+              <Spinner />
+            ) : (
+              <>
+                <ProjectForm
+                  product={product}
+                  startOpen={!projects.projects.length}
+                />
+                {Boolean(projects.projects.length) && (
```
wlonk: You can probably just let this get implicitly coerced to boolean.

src/js/store/projects/selectors.ts
```diff
@@ -0,0 +1,25 @@
...
+export const selectProjectsByProduct = createSelector(
+  [selectProjectState, selectProduct],
+  (
+    projects: ProjectsState,
+    product: Product | null | undefined,
+  ): ProjectsByProductState | undefined => {
+    if (product) {
+      return projects[product.id];
+    }
+    return undefined;
```
wlonk: Do we want to use `undefined` here or `null`? To my mind they mean slightly different things; we've tended to use `null` for "there is no value" and `undefined` for "there is no value, yet".

src/js/components/products/detail.tsx
duggiemitchell: @wlonk It's optional and it basically tells React: I only want to run this function if the values in this array have changed.

src/js/components/products/detail.tsx
jgerigmeyer: We could, if the code were adjusted to `{ projects.projects.length ? (...) : null}`. That is, using a ternary operator instead of a logical and operator. I don't have strong preferences on which is cleaner/clearer.

src/js/components/products/detail.tsx
jgerigmeyer: Technically optional, but strongly advised: https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

jest.config.js
jgerigmeyer: Hm, I could see the argument there for `index.tsx` at least. It's not a change in this PR, but I'll consider how difficult that would be.

test/js/components/products/detail.test.jsx
```diff
@@ -98,4 +118,98 @@ describe('<ProductList />', () => {
...
+              notFound: [],
+              fetched: true,
+            },
+          },
+        },
+      });
+
+      expect(queryByText('Load More')).toBeNull();
+    });
+  });
```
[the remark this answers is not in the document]
jgerigmeyer: @duggiemitchell I think this is correct as-is. It's testing aspects of the product-detail component, not project-detail.

metashare/api/models.py
```diff
@@ -132,12 +133,45 @@ def __str__(self):
...
+        return self.name
+
+    def save(self, *args, **kwargs):
+        if not self.branch_name:
+            self.branch_name = slugify(self.name)
+        super().save(*args, **kwargs)
+
+    class Meta:
+        ordering = ("-created_at", "name")
+        unique_together = (("name", "product"), ("branch_name", "product"))
```
jgerigmeyer: @wlonk If I try to create a new project with the same name but different capitalization as an existing project (on that product/repo), I get a `500` error:

```
IntegrityError at /api/projects/
duplicate key value violates unique constraint "api_project_branch_name_product_id_dd9a8e81_uniq"
DETAIL:  Key (branch_name, product_id)=(sample-project-54, 28) already exists.
```

PR #95, opened 2019-07-24
[task-create] Create tasks list

wlonk:
otter


metashare/api/serializers.py
```diff
@@ -63,3 +78,33 @@ class Meta:
...
+            "project",
+            "assignee",
+            "slug",
+            "old_slugs",
+        )
+        validators = (
+            UniqueTogetherValidator(
+                queryset=Task.objects.all(),
+                fields=("name", "project"),
+                message=_("A task with this name already exists."),
```
jgerigmeyer: @wlonk Similar to projects, I think we want to disallow names that are the same, case-insensitive. Right now I can add `task` and `Task` as separate items.

wlonk: Yeah. That ol' case.

stacyk:
[I want to create a list of tasks for a project with a name (and/or description?).](https://trello.com/c/sXr0nuDj/21-i-want-to-create-a-list-of-tasks-for-a-project-with-a-name-and-or-description)

metashare/api/validators.py
```diff
@@ -0,0 +1,30 @@
+from rest_framework.validators import UniqueTogetherValidator, qs_filter
+
+
+class CaseInsensitiveUniqueTogetherValidator(UniqueTogetherValidator):
+    def process_field_name(self, field_name):
+        """
+        Right now, we presume that certain names are string-y, and cane be
```
jgerigmeyer:
```suggestion
        Right now, we presume that certain names are string-y, and can be
```

src/sass/patterns/_tables.scss
```diff
@@ -0,0 +1,29 @@
+// Tables
+// ======
+
+.minimal-th {
```
mirisuzanne: This does look nicer than the default – is there SLDS precedent, or are we intentionally going off-script here?

jgerigmeyer: As far as I can tell we're just going off-script (cf https://www.lightningdesignsystem.com/components/data-tables/). @stacyk or @SondraE thoughts or reasoning?

stacyk: I believe this was designed before we knew we were using the datatable component. We originally had [expandable sections in a list](https://xd.adobe.com/view/755c3a4f-db06-4098-47a4-de2b8684fb8b-036d/) that didn't have a header row. Once it evolved into a table, I think either Sondra or I added the headers in bold but didn't include the header bg row. I do think it was a happy accident, but if it is a problem, we can remove this minimal-th to match more with SLDS

mirisuzanne: My instinct when using SLDS is generally to align whenever possible, unless that would look visually broken for some reason.

jgerigmeyer: @mirisuzanne @stacyk Changed to use standard SLDS table. Please re-review and cleanup as needed.

Screenshot 2019-08-12 17 21 42

PR #124, opened 2019-08-02
[fix-migrations] 🛢 Fix broken migrations

wlonk:
I had accidentally merged two migrations while working in #94. I've fixed that here, and it should help staging migrate forward seamlessly.

However, for local development environments, you will have to do the dance again; once you have pulled down this commit, run the following:

```
git checkout HEAD~1
./derrick migrate api 0001
git checkout fix-migrations
./derric migrate api
```

it me


wlonk:
To be clear, I reverted migration `0002` in `api` to match the name and content it had originally had, and pulled out the changes in `0002` and the content in `0003` into their own, new, `0003`.

PR #136, opened 2019-08-06
[devhub-connect] Add info to serializers about attached SF orgs

wlonk:
To make this work locally, run (with values from Keybase in place of `***`):

```
docker-compose run --rm web python manage.py populate_social_apps --sf-id *** --sf-secret ***
```

cat


wlonk:
@duggiemitchell @jgerigmeyer Once we add a frontend button to connect to a SF org, we can test out the Dev Hub requirement more effectively. Let me know if I can do anything to help, including especially serializer changes.

jgerigmeyer:
[I want to log into a Salesforce Org with a DevHub enabled.](https://trello.com/c/RNUFYqiF/16-i-want-to-log-into-a-salesforce-org-with-a-devhub-enabled)

metashare/api/tests/models.py
```diff
@@ -125,6 +139,7 @@ def test_full_org_type(self, user_factory, social_account_factory):
         user = user_factory(socialaccount_set=[])
         social_account_factory(
             user=user,
+            provider="salesforce-production",
```
davisagli: I see you're adding this a bunch of places; maybe it should be set as the factory's default?

metashare/multisalesforce/views.py
```diff
@@ -87,7 +87,7 @@ def parse_token(self, data):
 
         Called from OAuth2CallbackView"""
         data["access_token"] = fernet_encrypt(data["access_token"])
-        data["refresh_token"] = fernet_encrypt(data["refresh_token"])
+        # data["refresh_token"] = fernet_encrypt(data["refresh_token"])
```
davisagli: Why are we not storing the refresh token?

src/js/components/user/customDomainModal.tsx
```diff
@@ -53,30 +76,48 @@ const CustomDomainModal = ({ isOpen, toggleModal }: Props) => {
...
+            variant="brand"
+            style={{ width: `100%` }}
+            onClick={handleSubmit}
+          />
+          <Button
+            label={i18n.t('Log into Custom Org with DevHub')}
+            variant="outline-brand"
+            style={{ width: `100%` }}
+            onClick={() => setOrg('custom')}
+          />
```
davisagli: These button labels don't make sense. The login.salesforce.com endpoint will work for both production and Developer Edition orgs, and those same orgs can alternatively be accessed using the custom domain login endpoint.

metashare/api/models.py
```diff
@@ -76,28 +79,46 @@ def full_org_type(self):
...
+    def is_dev_hub_enabled(self):
+        token, _ = self.token
+        instance_url = self.salesforce_account.extra_data["instance_url"]
+        url = f"{instance_url}/services/data/v45.0/sobjects/ScratchOrgInfo"
+        resp = requests.get(url, headers={"Authorization": f"Bearer {token}"})
+        if resp.status_code == 200:
+            return True
+        if resp.status_code == 404:
+            return False
+        return None
```
davisagli: If we can determine the org is a sandbox or scratch org from the info that's stored in extra_data (see logic in metadeploy), we can short-circuit in that case and return False to avoid the API call.

davisagli: Looks like this property isn't used yet?

metashare/api/tests/models.py
wlonk: It's half the cases, actually; I'm gonna err on making `github` the default because it won't save me specification either way, and conceptually, a user should always have a github-provided social account, and only sometimes a salesforce-provided one.

metashare/multisalesforce/views.py
wlonk: Because I was debugging something and left that in there by mistake.

metashare/api/models.py
wlonk: It isn't used yet, no. I'd rather do it based on `extra_data` if we can, to avoid a nested HTTP call, but I don't know how.

jgerigmeyer:
@wlonk I'm still getting the front end cleaned up according to the new mockup, but now there's a dropdown in the header that you can use to test this functionally. I'm running into a few problems.

1. A `NoGitHubTokenError` because the `user_logged_in` signal attempts to call `user.refresh_repositories()` after SF connection.
2. It doesn't seem to handle both logins at the same time -- when I login with SF it effectively disconnects my user from GH, and vice versa.

wlonk:
@jgerigmeyer it looks like you're missing the `process=connect` key-value pair in the query string. Without that, it's making a separate login, not connecting another auth source to the same account.

src/js/components/user/connect.tsx
```diff
@@ -1,18 +1,153 @@
...
+    const val = url.trim();
+    if (!val) {
+      return;
+    }
+    const baseUrl = window.api_urls.salesforce_custom_login();
+    window.location.assign(
+      addUrlParams(baseUrl, {
+        custom_domain: val, // eslint-disable-line @typescript-eslint/camelcase
+        next: window.location.pathname,
+      }),
```
wlonk: Somewhere around here you should have `process: 'connect'`. But also for the production endpoint, not just the custom endpoint.

metashare/api/models.py
```diff
@@ -76,28 +79,50 @@ def full_org_type(self):
...
+                )
+            except InvalidToken:
+                return (None, None)
         return (None, None)
 
     @property
-    def social_account(self):
-        return self.socialaccount_set.first()
+    def salesforce_account(self):
+        return self.socialaccount_set.filter(provider="salesforce-production").first()
```
jgerigmeyer: @wlonk Ah, this explains why the `salesforce-custom` login isn't working currently.

wlonk: So we want to ensure that you only ever have one linked SF account at a time, from any provider, right? And we'll force disconnect the others. In which case this can be `provider__startswith="salesforce-"` and it'll work.

metashare/api/views.py
```diff
@@ -47,6 +47,17 @@ def post(self, request):
...
 
+class UserDisconnectSFView(CurrentUserObjectMixin, APIView):
+    model = User
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request):
+        user = self.get_object()
+        user.invalidate_salesforce_credentials()
+        serializer = FullUserSerializer(user)
+        return Response(serializer.data, status=status.HTTP_200_OK)
```
jgerigmeyer: @wlonk Note that I changed this to return the serialized User to avoid either an extra API call or making assumptions about the form of the updated data.

src/js/components/user/info.tsx
```diff
@@ -0,0 +1,196 @@
...
+      <Popover
+        align="bottom right"
+        heading={
+          <div className="slds-p-around_small">
+            <Icon category="utility" name="user" size="small" />
+            <span className="slds-p-left_small">{user.username}</span>
+            <div className="slds-m-left_large slds-p-left_small">
+              <Logout className="slds-text-body_regular" />
+            </div>
+          </div>
```
jgerigmeyer: @stacyk The combo of margins/padding to get this to line up is kinda ugly...

src/js/components/user/info.tsx
```diff
@@ -0,0 +1,196 @@
...
+        }
+        body={
+          <div className="slds-p-around_small">
+            {user.valid_token_for ? (
+              <ConnectionInfo user={user} />
+            ) : (
+              <ConnectToSalesforce toggleModal={setModalOpen} />
```
duggiemitchell: @jgerigmeyer I want to make sure I'm not missing something.  When I connect to sf, it doesn't change to show that it's connected :thinking: 
Screenshot from 2019-08-15 11-23-01

jgerigmeyer: That... is strange. 🤔 I assume you have the latest pulled, etc?

I wonder if we're suppressing an error message somewhere. If you navigate to `/admin/` do you see any helpful messages? Do you have the socialaccounts created etc?

duggiemitchell: ah, yes it does.  There's an error there that reads: `The social account is already connected to a different account.
`

jgerigmeyer: Ah, I ran into that earlier, but Kit fixed it late yesterday. Delete your existing SF socialaccounts from the admin, and then try again.

jgerigmeyer: @wlonk This does imply that we're not exposing (or even logging) connection errors in a helpful way. Added a story to address that: https://trello.com/c/e1dkbOKV/73-if-there-was-an-error-connecting-to-a-sf-org-i-want-it-to-be-exposed-to-the-user-with-helpful-instructions

jgerigmeyer: @duggiemitchell Did changing your permissions on the SF org resolve this issue for you?

CONTRIBUTING.rst
```diff
@@ -165,12 +165,13 @@ For more detailed instructions and options, see the `VS Code documentation`_.
...
-can be found in the GitHub App, or if you're an OddBird you can find these
-values in the shared Keybase team folder (``metashare/prod.db``)::
+To setup OAuth integration, run the ``populate_social_apps`` management command.
+The values to use in place of ``AAA``, ``BBB``, ``XXX`` and ``YYY`` can be found
+in the GitHub App and Salesforce App, respectively, or if you're an OddBird you
+can find these values in the shared Keybase team folder
+(``metashare/prod.db``)::
 
-    $ docker-compose run --rm web python manage.py populate_social_apps --gh-id XXX --gh-secret YYY
+    $ docker-compose run --rm web python manage.py populate_social_apps --gh-id AAA --gh-secret BBB --sf-id XXX --sf-secret YYY
```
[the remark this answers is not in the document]
jgerigmeyer: So the docs are written for someone setting up the project from scratch, in which case this command (which includes both `--sf-*` and `--gh-*) is correct. Since you (presumably) already had the GH connection setup, you only needed to run the `--sf-*` command.

jgerigmeyer:
@stacyk Yeah, as mentioned in the demo I did after standup the other day, I intentionally did not follow the mockup exactly here. We only have one item in the dropdown ("Connect to Salesforce") instead of a heading and two options.

I have a slight preference for the larger text since there's no header, but I'm happy to defer to whatever you and @SondraE or @mirisuzanne prefer.

Screenshot 2019-08-16 08 14 53

---

Screenshot 2019-08-16 08 15 02

config/settings/base.py
```diff
@@ -295,13 +295,10 @@ def env(name, default=NoDefaultValue, type_=str):
 # WHITENOISE_ROOT = PROJECT_ROOT.joinpath(static_dir_root)
 
 SOCIALACCOUNT_PROVIDERS = {
-    "github": {"SCOPE": ["read:user", "user:email", "repo", "read:org"]}
+    "github": {"SCOPE": ["read:user", "user:email", "repo", "read:org"]},
+    "salesforce-production": {"SCOPE": ["web", "full", "refresh_token"]},
+    "salesforce-custom": {"SCOPE": ["web", "full", "refresh_token"]},
 }
```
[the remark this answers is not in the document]
jgerigmeyer: If you login to login.salesforce.com, take a look at the URL for your org. Mine is https://na53.lightning.force.com -- you can then use e.g. `na53` as your "custom domain" to test this.

src/js/components/user/info.tsx
```diff
@@ -0,0 +1,208 @@
...
+            />
+            <div className="slds-p-left_x-large">
+              <p>{user.username}</p>
+              <Logout className="slds-text-body_regular slds-m-top_xx-small" />
+            </div>
+          </div>
+        }
+        body={
+          <div className="slds-p-around_small">
+            {user.valid_token_for ? (
```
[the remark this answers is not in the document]
jgerigmeyer: @duggiemitchell What you're seeing is not a normal "connected account w/o devhub enabled". It really has nothing to do with devhub at all. You're trying to log into an account where you don't have full permissions, so it's a unique case.

I agree we should handle it better (see https://trello.com/c/e1dkbOKV), but this is an error with your specific org and connecting in the first place, not with devhub or this particular piece of code.

We don't want to base "connection" on whether or not we have a `sf_username` -- `valid_token_for` is the signifier we want, and in this example you are not properly connected.

src/js/components/user/info.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+const UserInfo = () => {
+  const user = useSelector(selectUserState);
+  const [modalOpen, setModalOpen] = useState(false);
+
+  return user ? (
+    <>
+      <Popover
+        align="bottom right"
+        heading={
```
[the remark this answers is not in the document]
jgerigmeyer: @mirisuzanne Fixed.

Screenshot 2019-08-20 10 53 05
Screenshot 2019-08-20 10 53 32

PR #138, opened 2019-08-08
[task-detail] View task details

duggiemitchell:
see: https://trello.com/c/n6THbFfG

23-Bunny-Picture


jgerigmeyer:
[I want a detail view for a specific task to see more information than shown in the listing view.](https://trello.com/c/n6THbFfG/59-i-want-a-detail-view-for-a-specific-task-to-see-more-information-than-shown-in-the-listing-view)

src/js/components/products/detail.tsx
```diff
@@ -75,130 +74,82 @@ const ProductDetail = (props: RouteComponentProps) => {
...
-                    ) : null}
-                  </>
-                )}
+                <ul className="slds-has-dividers_bottom">
+                  {projects.projects.map(project => (
+                    <ProjectListItem
+                      key={project.id}
+                      project={project}
+                      product={product}
+                    />
```
wlonk: At what point do we want to have list-components that we render, rather than having the enclosing `ul` just verbatim in the calling component? I don't think we need that here, but I want to make sure we have thoughts about when that's appropriate and when it's unnecessary.

src/js/components/projects/detail.tsx
```diff
@@ -45,112 +43,61 @@ const ProjectDetail = (props: RouteComponentProps) => {
...
+    return <FourOhFour />;
   }
 
   if (
     (productSlug && productSlug !== product.slug) ||
     (projectSlug && projectSlug !== project.slug)
   ) {
-    // Redirect to most recent product slug
+    // Redirect to most recent product/project slug
     return <Redirect to={routes.project_detail(product.slug, project.slug)} />;
```
wlonk: This is project-only, right? But the comment seems to say otherwise. Am I missing something?

src/js/components/projects/detail.tsx
```diff
@@ -45,112 +43,61 @@ const ProjectDetail = (props: RouteComponentProps) => {
...
+          <>
+            <h2 className="slds-text-heading_medium slds-p-bottom_medium">
+              {tasks.length ? (
+                <>
+                  {i18n.t('Tasks for')} {project.name}
+                </>
+              ) : (
+                <>
+                  {i18n.t('Add a Task for')} {project.name}
+                </>
```
wlonk: Does this all need to be contained in a null element? Or can we just inline the strings?

src/js/components/utils.tsx
```diff
@@ -36,6 +41,105 @@ import {
...
+        className="slds-p-horizontal_x-large
+          slds-p-top_x-small
+          ms-breadcrumb"
+      >
+        <BreadCrumb
+          trail={[
+            <Link to={routes.home()} key="home">
+              {i18n.t('Home')}
+            </Link>,
+          ].concat(
```
wlonk: Does `concat` return the concatenated arrays? Different than the interface I expected, but makes sense.

src/js/components/utils.tsx
```diff
@@ -36,6 +41,105 @@ import {
...
+        info={<RepoLink url={repoUrl} shortenGithub />}
+        onRenderControls={onRenderHeaderActions}
+      />
+      <div
+        className="slds-p-horizontal_x-large
+          slds-p-top_x-small
+          ms-breadcrumb"
+      >
+        <BreadCrumb
+          trail={[
```
wlonk: I don't like this interface, but I recognize that it's not under our control, so :(

src/js/components/products/detail.tsx
jgerigmeyer: At this point this is the only list-item we're using, so I haven't thought a lot about best points of abstraction yet.

src/js/components/projects/detail.tsx
jgerigmeyer: See the `if`... it redirects if _either_ the product or project slug is outdated. You're correct that it renders project-detail in either case.

duggiemitchell: @winnettjohnston yes this should be project only I think

src/js/components/projects/detail.tsx
jgerigmeyer: Because it's concatenating strings, there are two choices:

```js
`${i18n.t('Add a Task for')} ${project.name}`
```

or

```js
<>{i18n.t('Add a Task for')} {project.name}</>
```

I don't really have an opinion as to which is cleaner. 🤷‍♂

src/js/components/utils.tsx
jgerigmeyer: Yep.

> This method does not change the existing arrays, but instead returns a new array.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

src/js/components/projects/detail.tsx
wlonk: Nor I. Just curious.

src/js/components/utils.tsx
jgerigmeyer: What part of it don't you like? We can control how the array is constructed, but we don't have control over the format of the eventual array.

This was the quick-and-dirty way to start with a common item, then add custom ones. I'm open to alternative suggestions.

wlonk: I would love a breadcrumb that just accepted an array and formatted it smartly, but I get that breadcrumb doesn't and this is _fine_ given those constraints.

src/js/components/projects/detail.tsx
jgerigmeyer: I think it depends how you're parsing the comment. It is redirecting to the most recent product _and_ project slugs, but redirecting to the project-detail URL.

src/js/components/utils.tsx
```diff
@@ -36,6 +41,105 @@ import {
...
+          {children}
+        </div>
+        <div
+          className="slds-col
+            slds-size_1-of-1
+            slds-medium-size_1-of-3
+            slds-text-longform"
+        >
+          {!descriptionHasTitle && (
+            <h2 className="slds-text-heading_medium">{title}</h2>
```
mirisuzanne: I see that it's in the UX proposal, but it feels strange to me having the page title duplicated on the sidebar.

jgerigmeyer: I'm open to changes there, but don't think it needs to hold up the PR; merging. 👍

PR #142, opened 2019-08-16
[scratch-org-create] Create a scratch org for a task.

jgerigmeyer:
To run this locally, copy the `server.key` from Keybase into your local repo (`compose/web/`):

    $ cp "/Volumes/Keybase ($USER)/team/oddbird/metashare/jwt/server.key" compose/web/

TODO:

- [x] update all relevant fields on scratch org after creation (https://github.com/SFDO-Tooling/MetaShare/pull/142#discussion_r319181927 and comments on https://github.com/SFDO-Tooling/MetaShare/commit/4e18f100c5d550f7b23d7fc1d5edff832614fb6d)
- [x] handle scratch-org-creation errors, emitting WS event (https://github.com/SFDO-Tooling/MetaShare/pull/142#discussion_r319550462)
- [x] prefix project branch_name with setting from cumulusci.yml (https://github.com/SFDO-Tooling/MetaShare/pull/142#discussion_r319551018)
- [x] review Jonny's changes: https://github.com/SFDO-Tooling/MetaShare/pull/142/files/b44d4b4e5ed933df4c5be2daa2881781557a0abb..09b19590ed9d48d4c841883a93e8df95e411b0d7
- [x] after org is created, run dev/qa org flow

otter


jgerigmeyer:
[I want to create a scratch org for a specific task.](https://trello.com/c/CPD7ls9L/23-i-want-to-create-a-scratch-org-for-a-specific-task)

metashare/api/models.py
```diff
@@ -218,6 +222,29 @@ class Meta:
...
+class ScratchOrg(mixins.HashIdMixin, mixins.TimestampsMixin, models.Model):
+    task = models.ForeignKey(Task, on_delete=models.PROTECT)
+    org_type = StringField(choices=SCRATCH_ORG_TYPES)
+    owner = models.ForeignKey(User, on_delete=models.PROTECT)
+    last_modified_at = models.DateTimeField(null=True)
+    expires_at = models.DateTimeField()
+    latest_commit = StringField(blank=True)
+    latest_commit_url = models.URLField(blank=True)
+    url = models.URLField()
+    has_changes = models.BooleanField(default=False)
```
jgerigmeyer: @wlonk For create, can we make `owner`, `expires_at`, and `url` optional?

wlonk: Yeah; `owner` should actually get set automatically, too.

metashare/api/models.py
```diff
@@ -213,11 +213,42 @@ class Task(mixins.HashIdMixin, mixins.TimestampsMixin, SlugMixin, models.Model):
...
 
+    def __str__(self):
+        return self.name
+
     class Meta:
         ordering = ("-created_at", "name")
         unique_together = (("name", "project"),)
 
 
+SCRATCH_ORG_TYPES = Choices("Dev", "QA")
```
jgerigmeyer: @wlonk Should these (and also the `ORG_TYPES` choices) be translatable on the back-end?

jgerigmeyer: That is, I'm trying to decide whether these values should be directly displayed on the front end (in which case they should probably be translatable on the back end), or should be hardcoded as display types for the front end (and translatable on the front end).

Dockerfile
```diff
@@ -40,6 +40,5 @@ RUN yarn install --check-files
 # if we override settings at runtime to be the local-dev settings, we'll build
 # and use local-dev assets.
 RUN yarn prod
-RUN python manage.py collectstatic --noinput
```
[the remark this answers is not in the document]
wlonk: Because it appears that the `--env` or `--env-file` flags implicitly passed in by `docker-compose` from the configuration yaml aren't available except at the `CMD` step. So the Django settings are wrong at this stage, and fail. But as I write this, I realize that, since `collectstatic` doesn't require a GitHub API key, I could just set a default on it in the settings and let that get overridden at the point of actual server start. Making that change now.

metashare/api/models.py
wlonk: I'm not sure; I could go either way. If I imagine this API being used by other frontends than ours, I'd rather it _not_ translate these strings, but that may be a fantasy anyway, so I'm not sure.

jgerigmeyer: Yeah, that makes sense to me. I tend to think the front end should be in control of display values, so I'll follow that here and make then translatable on the front end. I think it just threw me off that these were capitalized -- I expected `dev` and `qa` instead of `Dev` and `QA` -- but I'm not sure why.

metashare/consumers.py
```diff
@@ -13,7 +13,7 @@
 Request = namedtuple("Request", "user")
 
 
-KNOWN_MODELS = {"user"}
+KNOWN_MODELS = {"user", "scratchorg"}
```
jgerigmeyer: It'd be nice if we could make this `scratch_org`, but I found a workaround.

jgerigmeyer: In addition to `model` and `id`, I'm now also passing in `action: 'SUBSCRIBE' | 'UNSUBSCRIBE'`, which currently causes the `is_valid` check to fail.

metashare/api/models.py
```diff
@@ -213,11 +215,58 @@ class Task(mixins.HashIdMixin, mixins.TimestampsMixin, SlugMixin, models.Model):
...
```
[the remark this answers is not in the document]
jgerigmeyer: Created a new story: https://trello.com/c/OYUy81kP

metashare/api/models.py
```diff
@@ -210,14 +220,82 @@ class Task(mixins.HashIdMixin, mixins.TimestampsMixin, SlugMixin, models.Model):
...
+        return self.name
+
+    def save(self, *args, **kwargs):
+        if not self.branch_name:
+            self.branch_name = slugify(self.name)
+        super().save(*args, **kwargs)
+
     class Meta:
         ordering = ("-created_at", "name")
         unique_together = (("name", "project"),)
```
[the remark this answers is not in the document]
wlonk: I've been thinking about that, and I think that what we'll want to do instead is let GitHub be that source of truth. I've included logic to try branch names until one works, and I will soon include logic to start branch names at null and derive them from slugifying the title.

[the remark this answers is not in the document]
jgerigmeyer: Works for me. Might as well do the same for projects then.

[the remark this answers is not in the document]
wlonk: Already have.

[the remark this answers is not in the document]
jgerigmeyer: I meant removing the `unique_together` for `("branch_name", "repository")` on projects.

metashare/api/jobs.py
```diff
@@ -0,0 +1,53 @@
...
+    user, repo = extract_user_and_repo(repo_url)
+    repository = gh.repository(user, repo)
+    # Make project branch:
+    repository.create_branch_ref(
+        project_branch_name, repository.branch(repository.default_branch).latest_sha()
+    )
+    # Make task branch:
+    repository.create_branch_ref(
+        task_branch_name, repository.branch(project_branch_name).latest_sha()
+    )
```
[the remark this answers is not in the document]
wlonk: C'mon, in Python, we say `not`, we don't use illegible sigils! Except for function decorators.

metashare/api/models.py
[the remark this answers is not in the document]
jgerigmeyer: I see this is resolved, but it looks like the `unique_together` is still there?

metashare/api/models.py
```diff
@@ -204,20 +223,89 @@ class TaskSlug(AbstractSlug):
...
+    def subscribable_by(self, user):  # pragma: nocover
+        # TODO: revisit this?
+        return True
+
+    def save(self, *args, **kwargs):
+        create_remote_resources = self.id is None
+        super().save(*args, **kwargs)
+
+        if create_remote_resources:
+            self.create_remote_resources()
```
jgerigmeyer: It no longer seems like this is happening out-of-band. That is, the `POST` is just pending until the entire job finishes, then returns with a `201`.

jgerigmeyer: Fixed in https://github.com/SFDO-Tooling/MetaShare/pull/142/commits/ab2670b6553a54cb0420c12df09adbedd47593ab

metashare/api/models.py
```diff
@@ -190,34 +201,110 @@ def __str__(self):
...
+    def create_remote_resources(self):
+        from .jobs import create_branches_on_github_then_create_scratch_org_job
+
+        create_branches_on_github_then_create_scratch_org_job.delay(
+            project=self.task.project,
+            repo_url=self.task.project.repository.repo_url,
+            scratch_org=self,
+            task=self.task,
+            user=self.owner,
+        )
```
jgerigmeyer: If this job fails, trigger a WS event:

```
{
  "type": "SCRATCH_ORG_PROVISION_FAILED",
  "payload": {
    "error": "This is the error message.",
    "model": ScratchOrgSerializer(self).data,
  }
}
```

metashare/api/jobs.py
```diff
@@ -0,0 +1,96 @@
...
+            branch_name = f"{new_branch}-{counter}"
+
+
+def create_branches_on_github(*, user, repo_url, project, task):
+    gh = login(token=user.gh_token)
+    owner, repo = extract_owner_and_repo(repo_url)
+    repository = gh.repository(owner, repo)
+
+    # Make project branch:
+    project_branch_name = slugify(project.name)
```
jgerigmeyer: This should also get a prefix (defaults to "feature/"), pulled from `cumulusci.yml` on the repo.

metashare/api/jobs.py
```diff
@@ -30,8 +30,8 @@ class MetaDeployCCI(BaseCumulusCI):
 
 def create_scratch_org(scratch_org, *, user, repo_url, commit_ish):
     # We will eventually use scratch_org and user, but not yet.
-    make_scratch_org(user, repo_url, commit_ish)
-    scratch_org.url = "https://example.com"
+    url = make_scratch_org(user, repo_url, commit_ish)
+    scratch_org.url = url
```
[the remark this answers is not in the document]
wlonk: Where do we get `latest_commit` and `latest_commit_url` from? Put another way, what are those values? Values from GitHub? I have put in something, but I wanna leave this question here as a "please check my reasoning".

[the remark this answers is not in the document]
jgerigmeyer: Initially they are values from GH matching the `latest_sha` or `commit_ish` used to create the scratch org.

When changes are captured from an org and become a new commit, that commit then becomes the `latest_commit`.

metashare/api/jobs.py
wlonk: This turns out to occasion a valuable and worthwhile but large-ish refactor. I'm hoisting the GitHub checkout up above each of the other tasks, so it's available at this point where we need it, too, without repeatedly doing the checkout.

metashare/api/jobs.py
```diff
@@ -100,5 +92,5 @@ def create_branches_on_github_then_create_scratch_org_job(
         scratch_org=kwargs["scratch_org"],
         user=kwargs["user"],
         repo_url=kwargs["repo_url"],
-        commit_ish=kwargs["commit_ish"],
+        commit_ish=kwargs["task"].branch_name,
```
wlonk: I tried to actually deconstruct all the arguments, to make it possible to test without constructing weird objects with values on them (which is harder in Python than in JS), hence the explicit `commit_ish` kwarg, but I've unbundled the kwargs at this point, so.

metashare/api/serializers.py
```diff
@@ -91,6 +91,7 @@ class Meta:
     def get_branch_url(self, obj) -> Optional[str]:
         if obj.branch_name:
             return f"{obj.repository.repo_url}/tree/{obj.branch_name}"
+        return None
```
wlonk: I was actually discussing this with some friends, and was going to put in this change. I was trying out type annotation to indicate that the `None` branch was expected, but I think that including an explicit `None` is ultimately better.

metashare/api/jobs.py
```diff
@@ -0,0 +1,151 @@
...
+    gh = login(token=user.gh_token)
+    owner, repo = extract_owner_and_repo(repo_url)
+    repository = gh.repository(owner, repo)
+    branch = repository.branch(commit_ish)
+    latest_commit = branch.latest_sha()
+    # We can't use the urlt objects in repository because they build API urls:
+    latest_commit_url = f"{repository.html_url}/commit/{latest_commit}"
+
+    with report_errors_on(scratch_org):
+        org = ScratchOrgConfig({"config_file": "orgs/dev.json", "scratch": True}, "dev")
```
jgerigmeyer: @wlonk Should this depend upon the POSTed `org_type` instead of hard-coding `dev`?

davisagli: The `dev` org config should generally be fine for both dev and QA use cases.

metashare/api/jobs.py
```diff
@@ -0,0 +1,168 @@
...
+        org.create_org()
+        scratch_org.url = org.instance_url
+        scratch_org.last_modified_at = now()
```
jgerigmeyer: I'm not sure this needs to hold up merge, but I think `last_modified_at` and `latest_commit` need some more discussion. The layout makes it look like they are related, but functionally they aren't.

1. `latest_commit` is the latest commit when the scratch org is created, or the latest commit captured from the scratch org via MetaShare. Is `last_modified_at` the datetime of `latest_commit`, or the datetime of the latest change to the scratch org (which may not yet be captured)?

2. Related (but again should not hold up this PR), what happens if a task branch is modified (e.g. pushed to) outside of MetaShare? Then we get a case where the scratch org is out of date from the branch, which could cause conflicts trying to capture changes from the org.

metashare/api/jobs.py
```diff
@@ -0,0 +1,168 @@
...
+    message = {
+        "type": "SCRATCH_ORG_PROVISION_FAILED",
+        "payload": {"message": message, "model": ScratchOrgSerializer(instance).data},
+    }
+    model_name = instance._meta.model_name
+    id = str(instance.id)
+    group_name = "{model}.{id}".format(model=model_name, id=id)
+    channel_layer = get_channel_layer()
+    sent_message = {"type": "notify", "group": group_name, "content": message}
+    async_to_sync(channel_layer.group_send(group_name, sent_message))
```
jgerigmeyer: Is there a reason not to use `push_message_about_instance` here?

jgerigmeyer: @wlonk I hit an error creating a scratch org, but it was not exposed at all. Console:

```
/app/metashare/api/jobs.py:36: RuntimeWarning: coroutine 'RedisChannelLayer.group_send' was never awaited
  async_to_sync(channel_layer.group_send(group_name, sent_message))
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

```
Traceback (most recent call last):
  File "/app/metashare/api/jobs.py", line 42, in report_errors_on
    yield
  File "/app/metashare/api/jobs.py", line 86, in create_scratch_org
    org.create_org()
  File "/usr/local/lib/python3.7/site-packages/cumulusci/core/config/ScratchOrgConfig.py", line 206, in create_org
    raise ScratchOrgException(message)
cumulusci.core.exceptions.ScratchOrgException: Failed to create scratch org: 

ERROR running force:org:create:  The signup request failed because this organization has reached its active scratch org limit

id=no-job-id at=ERROR time="2019-09-04 14:04:52.361339" tag=external module=worker msg="cumulusci.core.exceptions.ScratchOrgException: Failed to create scratch org: 

ERROR running force:org:create:  The signup request failed because this organization has reached its active scratch org limit
Traceback (most recent call last):
  File \"/usr/local/lib/python3.7/site-packages/rq/worker.py\", line 822, in perform_job
    rv = job.perform()
  File \"/app/metashare/rq_worker.py\", line 12, in wrapper
    return orig(self)
  File \"/usr/local/lib/python3.7/site-packages/rq/job.py\", line 605, in perform
    self._result = self._execute()
  File \"/usr/local/lib/python3.7/site-packages/rq/job.py\", line 611, in _execute
    return self.func(*self.args, **self.kwargs)
  File \"/app/metashare/api/jobs.py\", line 167, in create_branches_on_github_then_create_scratch_org_job
    scratch_org=scratch_org, user=user, repo_url=repo_url, commit_ish=commit_ish
  File \"/app/metashare/api/jobs.py\", line 86, in create_scratch_org
    org.create_org()
  File \"/usr/local/lib/python3.7/site-packages/cumulusci/core/config/ScratchOrgConfig.py\", line 206, in create_org
    raise ScratchOrgException(message)
cumulusci.core.exceptions.ScratchOrgException: Failed to create scratch org: 

ERROR running force:org:create:  The signup request failed because this organization has reached its active scratch org limit
"
```

metashare/api/jobs.py
```diff
@@ -0,0 +1,168 @@
...
+
+@contextlib.contextmanager
+def report_errors_on(scratch_org):
+    try:
+        yield
+    except Exception as e:
+        report_scratch_org_error(scratch_org, str(e))
+        tb = traceback.format_exc()
+        logger.error(tb)
+        raise
```
jgerigmeyer: On an error, the scratch org DB model should be deleted (or at least flagged and omitted from future API calls).

metashare/api/jobs.py
wlonk: I'm getting that error over the websocket now. Can you confirm?

metashare/api/jobs.py
jgerigmeyer: Added a note to discuss as part of https://trello.com/c/8uR5ZTuC.

metashare/api/models.py
```diff
@@ -189,35 +200,119 @@ class Project(mixins.HashIdMixin, mixins.TimestampsMixin, SlugMixin, models.Mode
     def __str__(self):
         return self.name
 
+    def subscribable_by(self, user):  # pragma: nocover
+        # TODO: revisit this?
+        return True
```
jgerigmeyer: Right now this is only being used to live-update the `branch_url` while creating a scratch org... but I think it's okay?

metashare/api/models.py
```diff
@@ -189,35 +200,119 @@ class Project(mixins.HashIdMixin, mixins.TimestampsMixin, SlugMixin, models.Mode
...
+    branch_name = models.SlugField(max_length=50, null=True)
 
     slug_class = TaskSlug
 
+    def __str__(self):
+        return self.name
+
+    def subscribable_by(self, user):  # pragma: nocover
+        # TODO: revisit this?
+        return True
```
jgerigmeyer: Right now this is only being used to live-update the `branch_url` while creating a scratch org... but I think it's okay?

metashare/api/models.py
```diff
@@ -189,35 +200,119 @@ class Project(mixins.HashIdMixin, mixins.TimestampsMixin, SlugMixin, models.Mode
...
+    last_modified_at = models.DateTimeField(null=True)
+    expires_at = models.DateTimeField(null=True)
+    latest_commit = StringField(blank=True)
+    latest_commit_url = models.URLField(blank=True)
+    url = models.URLField(null=True)
+    has_changes = models.BooleanField(default=False)
+
+    def subscribable_by(self, user):  # pragma: nocover
+        # TODO: revisit this?
+        return True
```
jgerigmeyer: Probably okay for now? Any user subscribes to this if they load the page while the org is still being created.

jgerigmeyer:
@mirisuzanne 

>     * When I get the "devhub not enabled" modal, and click "check again", the modal disappears, and it's not clear what happened. Would it be possible to see feedback/results there, and an option to continue if the check is successful?

Hm... Are you sure you clicked "check again", not "Disconnect from Salesforce"? I am not able to reproduce with "check again". But yes, we can revisit the interactions there. https://trello.com/c/820XvLiR

>     * I'm not entirely sure where it is in the code, but the GitHub-branch button-inside-a-link is not valid html. I think the button is just being used for stylistic reasons? Styling links as buttons has also been falling out of favor for a11y reasons (the keyboard interactions are different), but I don't think it's a blocker for users.

Fixed and pushed.

>     * I wonder if there should be _some_ visual representation of "status" (without relying on the "warning" icon that was there initially)? It seems like an important line, but disappears in the box.

Is it an important line though? It will cause a big "Capture Changes" button to also appear on the page, which is really the important part. I'd vote for revisiting this later once we have more of the features in place.

docker-compose.yml
```diff
@@ -13,19 +13,20 @@ services:
...
     volumes:
       - .:/app
       - /app/node_modules
-      - ~/.gitconfig:/root/.gitconfig
     ports:
       - '0.0.0.0:8080:8080'
-        # Open this port to attach to remote debuggers in the Python process,
-        # if need-be:
-      - '0.0.0.0:33284:33284'  # Spells "DEBUG" in phone-number
+      - '0.0.0.0:1717:1717'  # For SFDX JWT Auth
```
jgerigmeyer: @wlonk So is this not actually necessary?

wlonk: At this moment I'm unsure, hold off until we can confirm with @davisagli .

davisagli: You need port 1717 open for completing the web server oauth flow, but shouldn't need it for the JWT flow.

metashare/api/jobs.py
```diff
@@ -73,9 +73,10 @@ def create_scratch_org(*, scratch_org, user, repo_url, commit_ish):
     owner, repo = extract_owner_and_repo(repo_url)
     repository = gh.repository(owner, repo)
     branch = repository.branch(commit_ish)
-    latest_commit = branch.latest_sha()
-    # We can't use the url objects in repository because they build API urls:
-    latest_commit_url = f"{repository.html_url}/commit/{latest_commit}"
+    commit = branch.commit
+    latest_commit = commit.sha
+    latest_commit_url = commit.html_url
+    latest_commit_at = commit.commit.author.get("date", None)
```
jgerigmeyer: @wlonk Happy for your review of these changes 🙏

metashare/api/jobs.py
```diff
@@ -0,0 +1,191 @@
...
+    cci = MetaShareCCI(
+        repo_root=repo_root,
+        repo_name="Repo Name",  # A placeholder, not used.
+        repo_url=repo_url,
+        repo_owner=owner,
+        repo_branch=repo_branch,
+        repo_commit=repository.branch(repo_branch).latest_sha(),
+    )
+    flow_config = cci.get_flow(cases[scratch_org.org_type])
+    flowrunner.FlowCoordinator(cci.project_config, flow_config).run(org_config)
```
jgerigmeyer:
```
id=no-job-id at=ERROR time="2019-09-06 15:00:51.248595" tag=external module=worker msg="AttributeError: 'list' object has no attribute 'values'
Traceback (most recent call last):
  File \"/usr/local/lib/python3.7/site-packages/rq/worker.py\", line 822, in perform_job
    rv = job.perform()
  File \"/app/metashare/rq_worker.py\", line 12, in wrapper
    return orig(self)
  File \"/usr/local/lib/python3.7/site-packages/rq/job.py\", line 605, in perform
    self._result = self._execute()
  File \"/usr/local/lib/python3.7/site-packages/rq/job.py\", line 611, in _execute
    return self.func(*self.args, **self.kwargs)
  File \"/app/metashare/api/jobs.py\", line 185, in create_branches_on_github_then_create_scratch_org
    org_config=org_config,
  File \"/app/metashare/api/jobs.py\", line 162, in run_appropriate_flow
    flowrunner.FlowCoordinator(cci.project_config, flow_config).run(org_config)
  File \"/usr/local/lib/python3.7/site-packages/cumulusci/core/flowrunner.py\", line 274, in __init__
    self.steps = self._init_steps()  # type: List[StepSpec]
  File \"/usr/local/lib/python3.7/site-packages/cumulusci/core/flowrunner.py\", line 421, in _init_steps
    self._check_infinite_flows(config_steps)
  File \"/usr/local/lib/python3.7/site-packages/cumulusci/core/flowrunner.py\", line 587, in _check_infinite_flows
    for step in steps.values():
AttributeError: 'list' object has no attribute 'values'
"
```

metashare/api/jobs.py
```diff
@@ -0,0 +1,191 @@
...
+        # json to the model, and use it to re-hydrate the
+        # ScratchOrgConfig as-needed.
+        scratch_org.url = org_config.instance_url
+        scratch_org.last_modified_at = now()
+        scratch_org.expires_at = org_config.expires
+        scratch_org.latest_commit = latest_commit
+        scratch_org.latest_commit_url = latest_commit_url
+        scratch_org.latest_commit_at = latest_commit_at
+        scratch_org.config = org_config.config
+        scratch_org.save()
```
jgerigmeyer: This triggers the success WS event, but I think we want to wait until the `dev_org` flow is done too.

metashare/api/jobs.py
wlonk: Working on this with @davisagli .

metashare/api/jobs.py
wlonk: Thanks, fixed.

metashare/api/jobs.py
```diff
@@ -0,0 +1,196 @@
...
+            user=user,
+            repo_url=repo_url,
+            project=project,
+            task=task,
+            repo_root=repo_root,
+        )
+        org_config = create_scratch_org(
+            scratch_org=scratch_org, user=user, repo_url=repo_url, commit_ish=commit_ish
+        )
+        run_appropriate_flow(
```
jgerigmeyer: @wlonk I think this (and maybe even `create_branches_on_github`) should be within a `with report_errors_on(scratch_org):` block?

jgerigmeyer:
[Revisit interactions of "Dev Hub not enabled" modal.](https://trello.com/c/820XvLiR/82-revisit-interactions-of-dev-hub-not-enabled-modal)

jgerigmeyer:
[I want to see an expiration date for each scratch org. (Scratch orgs are configured to last 30 days.)](https://trello.com/c/wSmTL9NU/24-i-want-to-see-an-expiration-date-for-each-scratch-org-scratch-orgs-are-configured-to-last-30-days)

metashare/api/sf_run_flow.py
```diff
@@ -0,0 +1,326 @@
...
+                        SETTINGS_XML_t.format(settingsName=settings_name, values=values)
+                    )
+            with open("package.xml", "w") as f:
+                f.write(PACKAGE_XML)
+
+            # Construct a new OrgConfig because ScratchOrgConfig
+            # tries to use sfdx which we don't want now -- this is a
+            # total hack which I'll try to smooth over with some
+            # improvements in CumulusCI
+            org_config = construct_org_config(
```
wlonk: Hoist this up and rename to `refresh_access_token`.

metashare/api/sf_run_flow.py
```diff
@@ -0,0 +1,326 @@
...
+    *, login_url, org_result, scratch_org_config, sf_client_id, sf_client_secret
+):
+    oauth = SalesforceOAuth2(sf_client_id, sf_client_secret, SF_CALLBACK_URL, login_url)
+    auth_result = oauth.get_token(org_result["AuthCode"]).json()
+    scratch_org_config.config["access_token"] = scratch_org_config._scratch_info[
+        "access_token"
+    ] = auth_result["access_token"]
+
+
+def save_settings(
```
wlonk: Rename to `deploy_org_settings`.

metashare/api/jobs.py
```diff
@@ -0,0 +1,178 @@
...
+    branch = repository.branch(commit_ish)
+    commit = branch.commit
+    latest_commit = commit.sha
+    latest_commit_url = commit.html_url
+    latest_commit_at = commit.commit.author.get("date", None)
+
+    org_config = ScratchOrgConfig(
+        {"config_file": "orgs/dev.json", "scratch": True}, "dev"
+    )
+    org_config.create_org()
```
wlonk: Pull out these two lines and use the structure in sf_run_flows, to avoid using sfdx.

metashare/api/jobs.py
```diff
@@ -0,0 +1,178 @@
...
+    # We just need to access this property to pull down info from SF as
+    # a side-effect.
+    org_config.scratch_info
```
wlonk: We won't need this line if we use the sf_run_flows approach.

metashare/api/jobs.py
```diff
@@ -0,0 +1,178 @@
...
+        stack.enter_context(report_errors_on(scratch_org))
+
+        commit_ish = create_branches_on_github(
+            user=user,
+            repo_url=repo_url,
+            project=project,
+            task=task,
+            repo_root=repo_root,
+        )
+        create_scratch_org(
```
wlonk: Return org here, once it's made in sf_run_flows style. Then pass it in to run_appropriate_flow, which will then be redesigned to presume it gets passed the org and doesn't have to make it.

wlonk: Instead, I'm using the combined create-org and run-flow logic in `sf_run_flows`.

wlonk:
Everything should be in here now. If you want to do one more functional test @jgerigmeyer and make sure I didn't miss anything in my refactoring, and @davisagli if you wanna give a 👍 to the code, we can merge this.

metashare/api/jobs.py
```diff
@@ -163,13 +145,16 @@ def create_branches_on_github_then_create_scratch_org(
...
+        create_org_and_run_flow(
+            scratch_org,
+            user=user,
+            repo_url=repo_url,
+            repo_branch=commit_ish,
+            project_path=repo_root,
         )
 
+        project.save()
+        task.save()
```
jgerigmeyer: I think we may want to do these after `create_branches_on_github` but before `create_org_and_run_flow`?

metashare/api/sf_run_flow.py
```diff
@@ -265,62 +267,61 @@ def run_flow(*, repo_owner, repo_name, repo_branch, user, flow_name):
...
+    # Scratch org construction is done but we haven't run a flow
+    # yet. This is the point where you would want to serialize
+    # scratch_org_config.config to store in the database for use
+    # later. Then reconstitute by running construc_org_config
+    # ---
+
+    # Run flow (takes care of getting a new access token)
+    flow = cci.get_flow(flow_name)
+    with cd(project_path):
+        flow.run(org_config)
```
jgerigmeyer: Missing a `return` statement?

metashare/api/jobs.py
jgerigmeyer: Maybe not, actually. Not sure it really matters -- just seems like maybe we want to save right away once we've successfully created the branches on GH. ¯\_(ツ)_/¯

wlonk: I can see the argument for that (keeps things closer to being in guaranteed sync) but worst case if something fails, we'll make new branches and leave some stray branches on GitHub.

metashare/api/sf_run_flow.py
wlonk: Doesn't actually return anything.

wlonk: Yeah, I just figured that out and pushed the fix :)

jgerigmeyer: Right... but I think that's the problem? It's being called as if it returns `org_config`: https://github.com/SFDO-Tooling/MetaShare/pull/142/files#diff-e148a690b7578d0eded316df5a9c8939R115

metashare/api/jobs.py
jgerigmeyer: Sure -- I guess I'm not clear on the argument the other way.

metashare/api/jobs.py
```diff
@@ -0,0 +1,163 @@
...
+        repo_owner=owner,
+        repo_name=repo,
+        repo_branch=repo_branch,
+        user=user,
+        flow_name=cases[scratch_org.org_type],
+        project_path=project_path,
+    )
+    scratch_org.url = org_config.instance_url
+    scratch_org.last_modified_at = now()
+    scratch_org.expires_at = org_config.expires
```
jgerigmeyer: This does not seem to be doing anything? Every org I create has no `expires_at` set.

wlonk: We remade it as an Org, not a ScratchOrg, and I guess that means it doesn't have an expires anymore. I'll fix it.

metashare/api/jobs.py
wlonk: Changed.

metashare/api/sf_run_flow.py
```diff
@@ -0,0 +1,329 @@
...
+            # GitHub App
+            "GITHUB_APP_ID": "",
+            "GITHUB_APP_KEY": """""",
+        }
+    )
+
+    repo_url = f"https://github.com/{repo_owner}/{repo_name}"
+    org_name = "dev"
+    devhub_username = user.sf_username
+    email = user.email  # TODO: check that this is reliably right.
```
jgerigmeyer: Yeah, this seems like an odd assumption to me. I'd expect it to come from the active SF connection, not the (GitHub-created) `user.email`.

metashare/api/sf_run_flow.py
```diff
@@ -0,0 +1,329 @@
...
+    return org_config
+
+
+def create_org_and_run_flow(
+    *, repo_owner, repo_name, repo_branch, user, flow_name, project_path
+):
+    # TODO: Should this come from env, even if that mirrors the DB contents?
+    sa = SocialApp.objects.filter(provider__startswith="salesforce").first()
+    sf_client_id = sa.client_id
+    sf_client_secret = sa.secret
```
jgerigmeyer: It seems like these should come from the active SF connection, not the DB or env. That is, aren't they ideally unique to whatever SF org the user authenticated with?

metashare/api/sf_run_flow.py
```diff
@@ -0,0 +1,329 @@
...
+            "SFDX_HUB_KEY": SF_CLIENT_KEY,
+            # GitHub App
+            "GITHUB_APP_ID": "",
+            "GITHUB_APP_KEY": """""",
+        }
+    )
+
+    repo_url = f"https://github.com/{repo_owner}/{repo_name}"
+    org_name = "dev"
+    devhub_username = user.sf_username
```
jgerigmeyer: Should this get passed into `get_org_result`? The username for my newly-created scratch org is "test-vihaqr25lpee@example.com", where I'd expect my SF username.

PR #149, opened 2019-08-23
[task-capture] Capture changes for a task.

jgerigmeyer:
pygmy possum


jgerigmeyer:
[I want to capture changes for a task at any point in the process.](https://trello.com/c/8uR5ZTuC/36-i-want-to-capture-changes-for-a-task-at-any-point-in-the-process)

metashare/api/models.py
```diff
@@ -291,6 +291,7 @@ class ScratchOrg(mixins.HashIdMixin, mixins.TimestampsMixin, models.Model):
     has_changes = models.BooleanField(default=False)
     currently_refreshing_changes = models.BooleanField(default=False)
     config = JSONField(default=dict, encoder=DjangoJSONEncoder)
+    login_url = models.URLField(null=True)
```
jgerigmeyer:
```suggestion
    login_url = models.URLField(null=True, blank=True)
```

metashare/api/migrations/0019_scratchorg_login_url.py
```diff
@@ -0,0 +1,14 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [("api", "0018_allow_blank_fields")]
+
+    operations = [
+        migrations.AddField(
+            model_name="scratchorg", name="login_url", field=models.URLField(null=True)
```
jgerigmeyer:
```suggestion
            model_name="scratchorg", name="login_url", field=models.URLField(null=True, blank=True)
```

metashare/api/sf_org_changes.py
```diff
@@ -0,0 +1,59 @@
+from cumulusci.salesforce_api.utils import get_simple_salesforce_connection
+
+from .github_context import extract_owner_and_repo, local_github_checkout
+from .sf_run_flow import BaseCumulusCI, refresh_access_token
+
+
+def get_salesforce_connection(
```
davisagli: I would probably turn this into a method of the ScratchOrg model. All of the inputs can be derived from the ScratchOrg or its project.

org_name should be stored on the ScratchOrg when it's created and used from there, so that this is usable for qa orgs too.

The base_url update should be an optional parameter (so default is the standard REST API but you can pass the param if you want the tooling API)

metashare/api/sf_org_changes.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    conn.base_url += "tooling/"
+
+    return conn
+
+
+def sf_org_has_changes(*, scratch_org, user):
+    repo_url = scratch_org.task.project.repository.repo_url
+    repo_owner, repo_name = extract_owner_and_repo(repo_url)
+    with local_github_checkout(user, repo_url) as project_path:
```
davisagli: We should not need to check out the code in order to make an API call. Yes, I know get_simple_salesforce_connection wants a project_config. This is why earlier I was trying to guide you to creating a custom subclass of BaseProjectConfig (and a custom subclass of BaseCumulusCI to construct it) that looks up what it needs from the database instead of from cumulusci.yml

metashare/api/jobs.py
```diff
@@ -55,6 +55,19 @@ def report_errors_on_provision(scratch_org):
...
+def report_errors_on_check_changes(scratch_org):
+    try:
+        yield
+    except Exception as e:
+        async_to_sync(report_scratch_org_error)(
+            scratch_org, str(e), "SCRATCH_ORG_FETCH_CHANGES_FAILED"
+        )
+        tb = traceback.format_exc()
+        logger.error(tb)
+        raise
```
davisagli: Is this a separate function for testability, or what? Since it's used in exactly one place I'd find it more readable if it were just implemented inline there.

jgerigmeyer: I think this should also set `has_changes: false` -- since it represents the fact that we can't reliably tell whether the org currently has changes or not. Either it'll need to fire a `SCRATCH_ORG_UPDATE` event, or make that change before the `SCRATCH_ORG_FETCH_CHANGES_FAILED` payload is serialized.

metashare/api/jobs.py
```diff
@@ -192,6 +205,17 @@ def create_branches_on_github_then_create_scratch_org(
 )
 
 
+def check_if_changes_on_org(*, scratch_org, user):
+    with report_errors_on_check_changes(scratch_org):
+        scratch_org.has_changes = sf_org_changes.sf_org_has_changes(
+            scratch_org=scratch_org, user=user
+        )
```
jgerigmeyer: Maybe you're still intending to do this, but I think we discussed replacing the boolean `has_changes` with the actual nested `changes` object (or `null` if no changes).

https://trello.com/c/8uR5ZTuC/36-i-want-to-capture-changes-for-a-task-at-any-point-in-the-process#comment-5d8100b20ec7da142ac3360a

metashare/api/jobs.py
```diff
@@ -192,6 +205,17 @@ def create_branches_on_github_then_create_scratch_org(
...
+        scratch_org.save()
```
jgerigmeyer: This should use `mark_refreshing_changes` so that it sets `currently_refreshing_changes` and triggers WS events accordingly?

I'll note that I'm not convinced abstracting the `mark_refreshing_changes` logic is actually helpful, nor using a tracker on `currently_refreshing_changes` to trigger the `SCRATCH_ORG_UPDATE` event. Aside from the bug we discovered with the job-instantiated tracker not having the same context as the model-instantiated one, it seems this amount of abstraction will result in multiple unnecessary `scratch_org.save()` calls?

As I understand it, this is what we want:

- On every org `GET`, set `currently_refreshing_changes: True` in the model (and in the `GET` response), and trigger a job to check for changes.
- On success, set `currently_refreshing_changes: False` and `scratch_org.changes` to the changes. Trigger a `SCRATCH_ORG_UPDATE` event.
- On failure, set `currently_refreshing_changes: False` and `scratch_org.changes: None`. Trigger either a `SCRATCH_ORG_UPDATE` event or `SCRATCH_ORG_FETCH_CHANGES_FAILED` event.

jgerigmeyer: Also: I think we only want to do this for Dev orgs, not QA orgs.

metashare/api/sf_org_changes.py
wlonk: @davisagli where is `org_name` at creation time? I'm guessing somewhere in the `scratch_org_config`, but I am not sure. I've been going through the various bits of documentation [here](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs.htm) to no avail yet.

metashare/api/jobs.py
```diff
@@ -55,6 +55,19 @@ def report_errors_on_provision(scratch_org):
         raise
 
 
+@contextlib.contextmanager
+def report_errors_on_check_changes(scratch_org):
+    try:
+        yield
+    except Exception as e:
+        async_to_sync(report_scratch_org_error)(
```
jgerigmeyer: I've noticed that the errors from SF are in a particular format:

```
[{'message': 'The signup request failed because this organization has reached its active scratch org limit', 'errorCode': 'LIMIT_EXCEEDED', 'fields': []}]
```

Would it make sense for the WS event to include just `e[0].message` or whatever instead of the entire thing?

metashare/api/sf_org_changes.py
davisagli: Never mind, for our purposes it's always dev.

metashare/api/jobs.py
wlonk: Yep, getting there bit by bit; I can't capture _what the changes are_ yet.

metashare/api/sf_org_changes.py
wlonk: So I'm trying to encapsulate the logic that interacts with SF in its own modules, to make it easier to change when/if changes arrive in cumulusci. I'm happy to put a method on ScratchOrg that calls out to this, but I'd like to keep the meat of the logic in this module.

metashare/api/jobs.py
wlonk: @davisagli can we reliably expect to get just one error, and extract it this way? Or do we need to account for the possibility of multiple errors in the list.

davisagli: Given that https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/errorcodes.htm says "The response body _usually_ contains...", I would avoid assuming too much.

metashare/api/jobs.py
wlonk: It's separate at this point because I think entering multiple contexts with an `ExitStack` is more legible than entwining multiple layers of setup and cleanup in a single `try-except`.

metashare/api/jobs.py
jgerigmeyer: Okay... can we do this only if there's one error, and it has a `message` property?

wlonk: @jgerigmeyer I feel like that would be more annoying on both the front and back end, to have to special case the one-item-array case? It would be moderately straightforward on the backend, but then the frontend would have to detect whether it got an array or an object and change display logic accordingly, right?

jgerigmeyer: No, the front end is just getting a string as it is now. I just think if the majority of the time we're getting a one-item array with a `message` property, let's default to displaying that to users.

Doing that check on the back end seems simplest to me, but if you prefer sending an actual array instead of a string I could do the check on the front end. 🤷‍♂

wlonk: OK; I thought it'd be more annoying for you, and wanted to make sure it wasn't. I'm happy to do it on the backend.

metashare/api/views.py
```diff
@@ -97,6 +98,21 @@ class TaskViewSet(viewsets.ModelViewSet):
     filter_backends = (DjangoFilterBackend,)
     filterset_class = TaskFilter
 
+    @action(detail=True, methods=["POST"])
```
jgerigmeyer: I don't really care practically, but in some ways it really feels like this belongs on the `org`, not the `task`. 🤷‍♂

metashare/api/views.py
```diff
@@ -97,6 +98,21 @@ class TaskViewSet(viewsets.ModelViewSet):
...
+        bad_scratch_orgs = (ScratchOrg.DoesNotExist, ScratchOrg.MultipleObjectsReturned)
+        task = self.get_object()
+        try:
+            scratch_org = task.scratchorg_set.filter(
+                org_type=SCRATCH_ORG_TYPES.Dev
+            ).get()
+            commit_changes_from_org_job.delay(scratch_org, request.user)
+            return Response("", status=status.HTTP_202_ACCEPTED)
+        except bad_scratch_orgs:
+            return Response("", status=status.HTTP_422_UNPROCESSABLE_ENTITY)
```
jgerigmeyer: When I actually try to `POST` to a task-detail, I get an error:

```
Method "POST" not allowed.
```

metashare/api/views.py
```diff
@@ -123,3 +108,18 @@ class ScratchOrgViewSet(viewsets.ModelViewSet):
 
     def perform_destroy(self, instance):
         instance.queue_delete()
+
+    def retrieve(self, request, *args, **kwargs):
+        instance = self.get_object()
+        instance.get_unsaved_changes()
+        serializer = self.get_serializer(instance)
+        return Response(serializer.data)
```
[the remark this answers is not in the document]
jgerigmeyer: Also: this should only happen on a Dev org.

metashare/api/jobs.py
```diff
@@ -184,17 +195,88 @@ def create_branches_on_github_then_create_scratch_org(
             repo_branch=commit_ish,
             project_path=repo_root,
         )
+        get_unsaved_changes(scratch_org, should_save=False)
```
jgerigmeyer: @wlonk Before this runs (probably above in `create_org_and_run_flow`?), it needs to store the _initial_ set of `latest_revision_numbers` on the org. Otherwise the first time `get_unsaved_changes` is called here, it returns _all_ revisions as if they're unsaved.

metashare/api/sf_org_changes.py
```diff
@@ -0,0 +1,109 @@
...
+    return record_dict
+
+
+def compare_revisions(old_revision, new_revision):
+    ret = defaultdict(list)
+    for mt in new_revision.keys():
+        for mn in new_revision[mt].keys():
+            if new_revision[mt][mn] > old_revision.get(mt, {}).get(mn, -1):
+                ret[mt].append(mn)
+    return ret
```
jgerigmeyer: @wlonk @davisagli Is it expected for this to return uncaptured changes for a newly-created Dev org, or is that a bug somewhere? When I create a brand new Dev org from https://github.com/oddbird/MetaShare-Test/ I get 16 uncaptured changes right away:

```json
{
  "latest_revision_numbers": {
    "Layout": {
      "SampleObject__c-SampleObject Layout": 19,
      "Child_Sample_Object__c-Child Sample Object Layout": 18
    },
    "Profile": {
      "Admin": 40
    },
    "WebLink": {
      "SampleObject__c.Just_Testing": 36,
      "Campaign.ViewCampaignInfluenceReport": 35
    },
    "ApexPage": {
      "SamplePage": 13
    },
    "ListView": {
      "SampleObject__c.All": 39,
      "Account.Household_Accounts": 37,
      "Opportunity.Default_Opportunity_Pipeline": 38
    },
    "ApexClass": {
      "SampleClass": 7,
      "SamplePage_CTRL": 9,
      "SampleClass_TEST": 8,
      "SamplePage_CTRL_TEST": 10
    },
    "CustomTab": {
      "SampleObject__c": 21
    },
    "RecordType": {
      "Account.HH_Account": 30,
      "Account.Organization": 31,
      "Opportunity.NPSP_Default": 32
    },
    "ApexTrigger": {
      "SampleTrigger": 11
    },
    "CustomField": {
      "SampleObject__c.Sample_Field__c": 26,
      "Child_Sample_Object__c.Parent__c": 25
    },
    "QuickAction": {
      "SampleObject__c.New_Child": 20
    },
    "CustomObject": {
      "SampleObject__c": 28,
      "Child_Sample_Object__c": 27
    },
    "ApexComponent": {
      "SampleComponent": 12
    },
    "CompactLayout": {
      "Account.NPSP_Household_Account": 33,
      "Account.NPSP_Organization_Account": 34
    },
    "HomePageLayout": {
      "SampleHomePage": 24
    },
    "BusinessProcess": {
      "Opportunity.NPSP_Default": 29
    },
    "CustomApplication": {
      "CumulusCITest": 22
    },
    "HomePageComponent": {
      "SampleComponent": 23
    }
  },
  "unsaved_changes": {
    "Profile": [
      "Admin"
    ],
    "WebLink": [
      "Campaign.ViewCampaignInfluenceReport",
      "SampleObject__c.Just_Testing"
    ],
    "ListView": [
      "Opportunity.Default_Opportunity_Pipeline",
      "SampleObject__c.All",
      "Account.Household_Accounts"
    ],
    "RecordType": [
      "Account.HH_Account",
      "Account.Organization",
      "Opportunity.NPSP_Default"
    ],
    "CustomField": [
      "Child_Sample_Object__c.Parent__c",
      "SampleObject__c.Sample_Field__c"
    ],
    "CustomObject": [
      "Child_Sample_Object__c",
      "SampleObject__c"
    ],
    "CompactLayout": [
      "Account.NPSP_Household_Account",
      "Account.NPSP_Organization_Account"
    ],
    "BusinessProcess": [
      "Opportunity.NPSP_Default"
    ]
  }
}
```

metashare/api/jobs.py
wlonk: That's by design; the intent is to set `latest_revision_numbers` with whatever was set up by the flow at org-creation. But that does also mark it as having unsaved changes, doesn't it?

jgerigmeyer: Right; we need to separate those two for the initial run.

metashare/api/jobs.py
```diff
@@ -184,17 +195,88 @@ def create_branches_on_github_then_create_scratch_org(
...
 
 
 create_branches_on_github_then_create_scratch_org_job = job(
     create_branches_on_github_then_create_scratch_org
 )
 
 
+def get_unsaved_changes(scratch_org, should_save=True):
```
jgerigmeyer: Now that this function is being called multiple places to update the changes (e.g. after a commit), it is triggering many unnecessary `SCRATCH_ORG_UPDATED` WS events.

metashare/api/sf_org_changes.py
wlonk: I believe those shouldn't show up, so I've made it now initialize latest-changes after running the initial flow.

metashare/api/jobs.py
```diff
@@ -36,6 +36,9 @@ def mark_refreshing_changes(scratch_org):
     scratch_org.save()
```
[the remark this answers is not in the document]
wlonk: I'll dig into this.

wlonk:
@jgerigmeyer aside from the delayed websocket events, I think I've responded to all your feedback.

metashare/api/views.py
```diff
@@ -107,3 +109,50 @@ class ScratchOrgViewSet(viewsets.ModelViewSet):
...
+    def list(self, request, *args, **kwargs):
+        # XXX: This method is copied verbatim from
+        # rest_framework.mixins.RetrieveModelMixin, because I needed to
+        # insert the get_unsaved_changes line in the middle.
+        queryset = self.filter_queryset(self.get_queryset())
+
+        # XXX: I am apprehensive about the possibility of flooding the
+        # worker queues easily this way:
+        for instance in queryset.filter(org_type=SCRATCH_ORG_TYPES.Dev):
+            instance.get_unsaved_changes()
```
jgerigmeyer: @wlonk So. You're correct here. A few page-refreshes can very quickly lead to a backlog of essentially-duplicate jobs in the queue. Some thoughts (but I'm not sure how technically feasible they are):

- If there's already a job running for that scratch_org, I think we can skip this. Practically I think that includes a `delete`, `create`, `commit`, or another `update` job. That is, if an org is actively being provisioned, deleted, capturing-changes, or already updating-unsaved-changes, I don't think we want to queue another update for that same org.
- If there's not a convenient automated way to do that (e.g. "hey there's another job running for this org, let's bail"), I think we could effectively do it by only triggering this `instance.get_unsaved_changes()` in the following cases:
  - `instance.org_type == SCRATCH_ORG_TYPES.Dev`
  - `instance.url != None`
  - `instance.delete_queued_at == None`
  - `instance.currently_capturing_changes == False`
  - `instance.currently_refreshing_changes == False`

wlonk: Research indicates that the thing to do is a lock (perhaps in Redis, like we did for MetaDeploy), where we use some unique natural key (the args passed to the job, for instance) and set the lock when we enqueue it, and unset it when the job completes or after a timeout. (The timeout we can do in-redis, I think, with setting an expiration on the key.)

wlonk: But I think we can use your option 2 more simply so I'd rather do that.

jgerigmeyer:
@mirisuzanne To test this functionally, comment out L169 of `orgs/actions.ts` and uncomment lines 171-219 of the same file. That will _mostly_ let you see what it will look like in practice, with one exception -- the "capture changes" btn will disappear while loading, whereas in practice it will stay visible, disabled with a spinner and loading message.

jgerigmeyer:
To functionally test (i.e. make your org think it has unsaved changes):

- Comment out L162 of `metashare/api/jobs.py` (https://github.com/SFDO-Tooling/MetaShare/blob/task-capture/metashare/api/jobs.py#L162)
- Go to your new org in the MS admin, put this for “Unsaved changes” and save it:

```
{"Layout": ["Child_Sample_Object__c-Child Sample Object Layout", "SampleObject__c-SampleObject Layout"], "Profile": ["Admin"], "WebLink": ["Campaign.ViewCampaignInfluenceReport", "SampleObject__c.Just_Testing"], "ApexPage": ["SamplePage"], "ListView": ["Opportunity.Default_Opportunity_Pipeline", "SampleObject__c.All", "Account.Household_Accounts"], "ApexClass": ["SampleClass", "SampleClass_TEST", "SamplePage_CTRL", "SamplePage_CTRL_TEST"], "CustomTab": ["SampleObject__c"], "RecordType": ["Account.HH_Account", "Account.Organization", "Opportunity.NPSP_Default"], "ApexTrigger": ["SampleTrigger"], "CustomField": ["Child_Sample_Object__c.Parent__c", "SampleObject__c.Sample_Field__c"], "QuickAction": ["SampleObject__c.New_Child"], "CustomObject": ["Child_Sample_Object__c", "SampleObject__c"], "ApexComponent": ["SampleComponent"], "CompactLayout": ["Account.NPSP_Household_Account", "Account.NPSP_Organization_Account"], "HomePageLayout": ["SampleHomePage"], "BusinessProcess": ["Opportunity.NPSP_Default"], "CustomApplication": ["CumulusCITest"], "HomePageComponent": ["SampleComponent"]}
```

src/sass/patterns/_form-elements.scss
```diff
@@ -1,16 +1,47 @@
 // Form Elements
 // =============
 
-.vertical-separator {
-  border-right: 1px solid $color-border-separator-alt;
+.form-grid {
+  display: grid;
+  grid-template-columns: 1fr 25%;
+  flex-grow: 1;
+  width: calc(100% - 2.25rem);
```
mirisuzanne: The issues here are caused by using flex to achieve grid-like layouts. If you could make the `slds-accordion__summary` act as a grid container, rather than a flex container, you could eliminate both these lines. The only magic number you'd really need is margin/padding on the header row to get it the same width as accordion rows.

Another option is to blame `.slds-has-flexi-truncate` on the `h3` - which is allowing it to grow more than it should. That seems like a more complicated fix to me, though.

src/sass/patterns/_form-elements.scss
```diff
@@ -1,16 +1,47 @@
...
+        .slds-form-element__label {
+          color: $brand-text-link-active;
+        }
+      }
+    }
   }
 }
+
+.selectall-header {
+  margin-left: 18px; // 18px chevron icon placeholder
```
mirisuzanne: I think this is the best place to capture any magic numbers. If you get this row aligned with the others, and using the same grid - things should line up internally.

src/sass/patterns/_form-elements.scss
```diff
@@ -1,16 +1,47 @@
...
+}
+
+.changes-col {
+  margin-left: 3px; // magic number to help line up changes column
```
mirisuzanne: I'd guess that this means something is misaligned on the parent `.form-grid`

src/sass/patterns/_form-elements.scss
```diff
@@ -1,16 +1,47 @@
...
   height: 100%;
-  width: 1px;
 
-  .hide-separator + & {
-    display: none;
+  // using slds class since we can't access inner checkbox + label markup
+  .slds-checkbox__label {
+    &:hover {
+      .slds-form-element__label {
+        color: $brand-text-link-active;
```
mirisuzanne: seems weird that we would have to add this hover change - do they not have one built in?

src/js/components/orgs/capture.tsx
```diff
@@ -0,0 +1,313 @@
...
+                  noChildrenChecked = false;
+                  checkedChildren = checkedChildren + 1;
+                }
+              }
+              if (checkedChildren === children.length) {
+                allChildrenChecked = true;
+              }
+
+              return (
+                <Accordion key={groupName} className="light-bordered-row">
```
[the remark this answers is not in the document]
jgerigmeyer: We don't really have control about the `h3` around the `button` -- that's straight from SLDS (https://www.lightningdesignsystem.com/components/accordion/). But we are using the empty `span` as a workaround to mimic the Summary Detail component (https://www.lightningdesignsystem.com/components/summary-detail/) which is not yet implemented in React-SLDS (https://github.com/salesforce/design-system-react/issues/2231). Basically we had to decide whether the title would be a click target to expand the accordion, or a click target to select the checkbox. We went with the latter (moving it into the label), but the result was this empty `span` (until we get a real Summary Detail component that allows an empty summary title).

src/sass/patterns/_form-elements.scss
stacyk: I thought it was unusual too, but nothing is shown on the label nor checkbox (faux checkbox) and it is part of the click target: https://www.lightningdesignsystem.com/components/checkbox/

src/sass/patterns/_form-elements.scss
stacyk: I don't think I will be able to add a class to `slds-accordion__summary` but I can target our custom `data-form='task-capture'` and then the `slds-accordion__summary`

src/js/components/orgs/capture.tsx
```diff
@@ -0,0 +1,313 @@
...
+              ) => handleSelectGroup(groupName, checked);
+              let checkedChildren = 0;
+              let allChildrenChecked = false;
+              let noChildrenChecked = true;
+              for (const child of children) {
+                if (
+                  (inputs as Inputs).changes[groupName] &&
+                  (inputs as Inputs).changes[groupName].includes(child)
+                ) {
+                  noChildrenChecked = false;
```
wlonk: Can this just be `checkedChildren === 0`?

src/js/components/tasks/detail.tsx
```diff
@@ -115,6 +184,23 @@ const TaskDetail = (props: RouteComponentProps) => {
...
+    buttonText = (
+      <LabelWithSpinner
+        label={i18n.t('Capturing Selected Changes…')}
+        variant="inverse"
+      />
+    );
+  } else if (fetchingChanges || currentlyFetching) {
+    buttonText = (
+      <LabelWithSpinner
+        label={i18n.t('Checking for Uncaptured Changes…')}
```
wlonk: Do we want these things to be title-case?

src/js/store/orgs/reducer.ts
```diff
@@ -164,6 +193,21 @@ const reducer = (
       }
       return orgs;
     }
+    case 'SCRATCH_ORG_COMMIT_CHANGES_FAILED':
+    case 'GITHUB_CHANGES_COMMITTED': {
```
wlonk: Should these events be in the same quasi-namespace?

src/js/utils/helpers.ts
```diff
@@ -0,0 +1,41 @@
+import i18n from 'i18next';
```
wlonk: Maybe rename this file to `messageHelpers` or similar; it's oriented towards user-visible strings entirely.

src/js/components/tasks/detail.tsx
jgerigmeyer: I'd leave that up to @stacyk and @SondraE, but it does match the CTA buttons elsewhere through the app.

src/sass/patterns/_form-elements.scss
stacyk: the newest set of changes resolves this

stacyk:
> The UX / functionality looks great. Left a few other comments for consideration.

@mirisuzanne I added the grid styles and edited the padding classes to remove almost all magic numbers (except for the one that controls the changes column width. I think if we had subgrid, we wouldn't need it.

src/sass/patterns/_form-elements.scss
```diff
@@ -1,16 +1,52 @@
...
 
-.vertical-separator {
-  border-right: 1px solid $color-border-separator-alt;
+.form-grid {
+  display: grid;
+  // needs a non-percentage and non fr
+  //   for changes column to be consistent between row types
+  //   16vw is a magic number that works well in changes column
+  grid-template-columns: 1fr 16vw;
+}
```
jgerigmeyer: If this is specific to the task-capture / changes layout, should it be nested under `[data-form='task-capture']`?

src/js/store/orgs/reducer.ts
wlonk: No, for now.

jgerigmeyer: > No, for now.

I'm curious why. It seems like `SCRATCH_ORG_COMMIT_CHANGES` would make sense?

wlonk: I guess the frontend does care that it's about an org; I kept seeing it from the backend's perspective, but the event is consumed by the frontend, so.

metashare/api/jobs.py
```diff
@@ -88,15 +38,16 @@ def try_to_make_branch(repository, *, new_branch, base_branch):
                 raise
 
 
-def create_branches_on_github(*, user, repo_url, project, task, repo_root):
+def _create_branches_on_github(*, user, repo_url, project, task, repo_root):
     """
     Expects to be called in the context of a local github checkout.
     """
-    gh = login(token=user.gh_token)
+    gh = gh_given_user(user)
```
jgerigmeyer: @wlonk It looks like currently `user.gh_token` and `user.github_account` are not being used. We can probably use them (within `gh_given_user`) or just remove them?

src/sass/patterns/_form-elements.scss
mirisuzanne: Is there a reason we want to override the design system in this one specific case? I'd expect we either trust them on this, or override it globally?

src/sass/patterns/_form-elements.scss
mirisuzanne: only if this class is used other places where we don't want the behavior. But I think this isa new class created specifically for this. Could change the name to `task-capture-form` or  something if it needs more clarity, rather than nesting.

src/js/components/orgs/capture.tsx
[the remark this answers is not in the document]
jgerigmeyer: @mirisuzanne Do you have an alternative proposal here? Or requested changes?

PR #151, opened 2019-08-26
[scratch-org-delete] Delete a scratch org.

jgerigmeyer:
labradoodle


jgerigmeyer:
[I want to delete a scratch org.](https://trello.com/c/NvxDQ2BH/50-i-want-to-delete-a-scratch-org)

metashare/api/models.py
```diff
@@ -295,6 +296,18 @@ def save(self, *args, **kwargs):
...
+        if self.tracker.has_changed("currently_refreshing_changes"):
+            self.notify_refreshing_changes()
+
+    def queue_delete(self):
+        from .jobs import delete_scratch_org_job
+
+        delete_scratch_org_job.delay(self)
+
+    def delete(self, *args, **kwargs):
+        self.notify_deleted()
```
jgerigmeyer: This is now triggered by https://github.com/SFDO-Tooling/MetaShare/pull/151/files#diff-e148a690b7578d0eded316df5a9c8939R52 when a scratch-org-create action fails. That's not ideal, from the perspective of the front end at least, because it can sometimes (depending on WS race conditions) result in a "You have successfully deleted this scratch org" message when really the org just wasn't successfully created. Can we make this more specific somehow, tying the WS event more directly to the action it is describing?

metashare/api/models.py
```diff
@@ -313,6 +326,20 @@ def notify_has_url(self):
         message = {"type": "SCRATCH_ORG_PROVISIONED", "payload": payload}
         async_to_sync(push.push_message_about_instance)(self, message)
 
+    def notify_refreshing_changes(self):
+        from .serializers import ScratchOrgSerializer
+
+        payload = ScratchOrgSerializer(self).data
+        message = {"type": "SCRATCH_ORG_UPDATED", "payload": payload}
+        async_to_sync(push.push_message_about_instance)(self, message)
```
jgerigmeyer: Every time I create a scratch org and receive a `SCRATCH_ORG_PROVISIONED` event, I'm also receiving a `SCRATCH_ORG_UPDATED` with the exact same org payload.

metashare/api/views.py
```diff
@@ -102,3 +102,6 @@ class ScratchOrgViewSet(viewsets.ModelViewSet):
     queryset = ScratchOrg.objects.all()
     filter_backends = (DjangoFilterBackend,)
     filterset_class = ScratchOrgFilter
+
+    def perform_destroy(self, instance):
+        instance.queue_delete()
```
jgerigmeyer: When we start a deletion process, we should store that data somewhere on the model so that after a page-refresh the front end still knows that it's being deleted. The most obvious to me would be setting a `deletion_queued_at` timestamp on the model and saving it.

metashare/api/jobs.py
```diff
@@ -161,3 +172,11 @@ def create_branches_on_github_then_create_scratch_org(
 create_branches_on_github_then_create_scratch_org_job = job(
     create_branches_on_github_then_create_scratch_org
 )
+
+
+def delete_scratch_org(scratch_org):
+    sf_run_flow.delete_scratch_org(scratch_org)
```
jgerigmeyer: This should run inside an error handler (similar to `report_errors_on`). On a deletion error, we should:

- set `deletion_queued_at` back to `None` on the model, and save it (*not* deleting the model)
- then emit a WS event: `{ type: 'SCRATCH_ORG_DELETE_FAILED', payload: { message: <string>, model: <Org> } }` (this should happen after, so that `deletion_queued_at` is `null` in the serialized WS payload)

metashare/api/models.py
wlonk: Would you be OK with just the `SCRATCH_ORG_UPDATED` event, then? `PROVISIONED` doesn't have to be a distinct concept.

jgerigmeyer: It's not clear to me why this is happening -- what is causing `self.tracker.has_changed("currently_refreshing_changes")` to be truthy when the org is provisioned?

From my perspective, it's not ideal to combine these into one "updated" event, because we want to display a message to the user on successful org-creation but not every time the org model is updated.

wlonk: OK, I've been digging into this and I'm baffled; that isn't changed anywhere in the code as it stands. So I'm also not clear on why it's happening. I'll leave it as-is for now and try to step through the server's behavior.

jgerigmeyer: It would be nice to figure out, but also isn't actively breaking anything, so probably shouldn't hold up review/merge when this PR is ready.

metashare/api/jobs.py
```diff
@@ -20,28 +20,59 @@
...
 
 
+@contextlib.contextmanager
+def report_errors_on_delete(scratch_org):
+    try:
+        with report_errors_on(scratch_org, "SCRATCH_ORG_DELETE_FAILED"):
+            yield
+    except Exception:
+        scratch_org.delete_queued_at = None
+        scratch_org.save()
```
jgerigmeyer: It looks like this is actually happening _after_ the org is serialized for the `SCRATCH_ORG_DELETE_FAILED` event. That's okay if it's better/easier that way, or necessary -- just let me know because I'll need to adjust the JS to manually remove the `delete_queued_at` if that's the case.

wlonk: It's a little easier to decompose that way, but I think you're right that it's wrong to send it that way, so I'll restructure this to fix that.

metashare/api/models.py
jgerigmeyer: Seems like within the job a new `FieldTracker` is being instantiated, which means it loses track of which fields have already been saved (e.g. the "current" state). This causes new unnecessary events to be triggered on the `scratch_org.save()` call made from the job.

Added a task to https://trello.com/c/8uR5ZTuC.

metashare/api/sf_run_flow.py
```diff
@@ -304,3 +304,26 @@ def create_org_and_run_flow(
...
+    # Update environment with some keys CumulusCI uses
+    # TODO: Should this be in a context manager so as not to leak across
+    # runs?
+    os.environ.update(
+        {
+            # Salesforce connected app
+            "SFDX_CLIENT_ID": SF_CLIENT_ID,
+            "SFDX_HUB_KEY": SF_CLIENT_KEY,
+        }
+    )
```
davisagli: We shouldn't need to set these environment variables at runtime.

metashare/api/sf_run_flow.py
```diff
@@ -304,3 +304,26 @@ def create_org_and_run_flow(
...
+
+    devhub_username = scratch_org.owner.sf_username
+    org_id = scratch_org.config["org_id"]
+
+    devhub_api = get_devhub_api(devhub_username=devhub_username)
+    active_scratch_org_id = devhub_api.query(
+        f"SELECT Id FROM ActiveScratchOrg WHERE ScratchOrg='{org_id}'"
+    )["records"][0]["Id"]
```
davisagli: Need to handle the case where this returns no records because the scratch org was already deleted some other way.

metashare/api/sf_run_flow.py
wlonk: Do we need to set them up on L237, at the start of `create_org_and_run_flow`? I imagine not, and would love to remove them there too.

davisagli: Like I said in slack, I think we're bypassing the places in CumulusCI where they are used. But in order to be safe I would recommend setting these in the MetaShare environment outside of runtime (i.e. rename the existing SF_CLIENT_ID and SF_CLIENT_KEY env vars to SFDX_CLIENT_ID and SFDX_CLIENT_KEY)

Hmm, except CumulusCI doesn't know how to deal with the escaped newlines in the key.

wlonk: Which, unfortunately, are the only way Docker can inject them. It should work on Heroku, but it'll be a mess in local dev.

davisagli: We could update the environment to set SFDX_CLIENT_KEY based on the processed SF_CLIENT_KEY in the settings module at import time rather than doing it at runtime here. Let's wait until we discover that we have a real need for it though.

src/js/store/orgs/actions.ts
```diff
@@ -110,3 +132,127 @@ export const provisionFailed = ({
...
+        payload: { org, url, response },
+      });
+    }
+    // @@@ Mock out until API exists
+    // setTimeout(() => {
+    //   dispatch({
+    //     type: 'SCRATCH_ORG_UPDATED',
+    //     payload: { ...response },
+    //   });
+    // }, 3000);
```
wlonk: Can we clear out this comment?

src/js/utils/websockets.ts
```diff
@@ -74,10 +98,16 @@ export const getAction = (event: EventType) => {
...
+    case 'SCRATCH_ORG_DELETED':
+      return deleteOrg(event.payload);
+    case 'SCRATCH_ORG_DELETE_FAILED':
+      return deleteFailed(event.payload);
     case 'PROJECT_UPDATE':
       return updateProject(event.payload);
     case 'TASK_UPDATE':
       return updateTask(event.payload);
+    case 'SCRATCH_ORG_UPDATED':
+      return updateOrg(event.payload);
```
wlonk: Can we order these cases so that they're logically grouped?

src/js/store/orgs/actions.ts
jgerigmeyer: Nope, not until you make the `GET` endpoint actually check for updated changes. :)

I expect it'll get cleared in the next PR -- for now it's helpful in dev.

PR #175, opened 2019-09-13
Clean up scratch-org-create

wlonk:
Gary the Meowntaineering Cat


Dockerfile
```diff
@@ -17,12 +17,6 @@ ENV DJANGO_SECRET_KEY 'sample secret key'
 ENV DJANGO_SETTINGS_MODULE config.settings.production
 ENV SFDX_CLIENT_ID 3MVG9oNqAtcJCF.FGhrPHhh9wPn3TbvF6mcqZfijFj0Zc.I4J6N0YCfsW6qZ.XKVAluOAAijDKV0_B1t3ZI0Z
```
davisagli: no longer used, right?

metashare/api/sf_run_flow.py
```diff
@@ -267,6 +250,12 @@ def create_org_and_run_flow(
     devhub_username = user.sf_username
     email = user.email  # TODO: check that this is reliably right.
```
jgerigmeyer: This seems like an odd assumption to me. I'd expect it to come from the active SF connection, not the (GitHub-created) `user.email`.

config/settings/base.py
```diff
@@ -348,6 +348,12 @@ def env(name, default=NoDefaultValue, type_=str):
 }
 
 
+# SF client settings:
+SF_CALLBACK_URL = env("SF_CALLBACK_URL", default=None)
+SF_CLIENT_KEY = env("SF_CLIENT_KEY", default=None)
+SF_CLIENT_ID = env("SF_CALLBACK_URL", default=None)
+SF_CLIENT_SECRET = env("SF_CLIENT_SECRET", default=None)
```
jgerigmeyer: So these are hardcoded values from the env? It seems like they should come from the active SF connection of the current user, not the DB or env. That is, aren't they ideally unique to whatever SF org the user authenticated with?

metashare/api/sf_run_flow.py
```diff
@@ -267,6 +250,12 @@ def create_org_and_run_flow(
     devhub_username = user.sf_username
```
jgerigmeyer: Should this get passed into `get_org_result`? The username for my newly-created scratch org is "test-vihaqr25lpee@example.com", where I'd expect my SF username.

metashare/api/sf_run_flow.py
davisagli: We're deciding what email to request for the user that will be provisioned in the scratch org. Either one of these is a reasonable guess for something that is likely to reach the person who is creating the scratch org, but the one from github will work better in a scenario when multiple MS users are sharing the same dev hub user.

config/settings/base.py
davisagli: No, these are the parameters for the connected app which is global for the MS instance. A single installation of MetaShare uses a single connected app to identify itself to the Salesforce platform as it logs in as multiple users.

metashare/api/sf_run_flow.py
davisagli: This is the SF username to connect to the dev hub, not the username for the scratch org. test-vihaqr25lpee@example.com is the expected format for scratch org users

config/settings/base.py
jgerigmeyer: @wlonk Can we update the docs (and presumably `env.example`) with instructions for setting these? I'd assume the `populate_social_apps` command could just use these env vars instead of passed-in args, for SF at least.

wlonk: @jgerigmeyer Let me know if this latest push satisfies your request.

CONTRIBUTING.rst
```diff
@@ -36,6 +36,19 @@ Docker-based development
...
+   Set these environment variables::
+
+    SF_CALLBACK_URL='http://localhost:8080/accounts/salesforce-production/login/callback/'
+    SF_CLIENT_KEY='-----BEGIN RSA PRIVATE KEY-----[some key here]-----END RSA PRIVATE KEY-----'
+    SF_CLIENT_ID=...
+    SF_CLIENT_SECRET=...
+
+   The ``SF_CLIENT_ID`` and ``SF_CLIENT_SECRET`` values should be the
+   same as the values you will eventually put in the database when you
+   initialize the SocialApps. The value for ``SF_CLIENT_KEY`` can be
```
davisagli: Hmm, I don't see any instructions to run populate_social_apps in here?

For our other projects we have been intending to add a migration step that populates the social apps so that a separate command doesn't need to be run. Can you go ahead and do that for MetaShare?

wlonk: That's because the `populate_social_apps` instructions were already in the document. I've added to them.

I think that adding setup data to tables is a wrong thing to do via a Django migration; can you walk me through the rationale a bit? (In particular, as it's something that reasonably differs between deployments, including it in the migrations makes the set up brittle and over-specific, I think.)

davisagli: The rationale is that we always want them there, so it's frustrating to have an extra step which we sometimes forget to run.

The differences between deployments come down to the client id/secret which now default to the environment variables, right? Am I forgetting some scenario where we don't want them there?

wlonk: So you're saying that the migration should pull the values from the env? That leaves me feeling really uncomfortable, but I can't articulate why.

davisagli: Yeah. I mean really I'd like them to come from the env at runtime rather than migration time, but since that's not really how django-allauth works, this at least gets us a step closer. We still need to remember to update the values in the db if we change them in the environment, but this would at least remove one time that we have to do that.

jgerigmeyer: The GitHub values are not currently pulled from any env vars. I tend to agree with Kit on what seems cleaner here from a theoretical perspective, but I understand the practical convenience of adding a setup migration. It does seem like a separate PR for that change would be fine, and it doesn't need to hold up this merge.

wlonk: I think part of why I am uncomfortable with this is that it makes applying migrations non-repeatable; the resultant DB state will depend on a _very literal_ environmental factor. If you're OK with that, we can do it, but I think it's generally worth avoiding.

davisagli: What's so repeatable about migrations to begin with? Their behavior already generally depends on what data is in the db and changes over time as more migrations are added. So I'm not sure I understand what design goal you're aiming for.

wlonk: It's not important.

PR #181, opened 2019-09-19
[restrict-github] Refresh GitHub repositories out-of-band

wlonk:
white rhino

See https://trello.com/c/NfO93RKQ/64-only-query-github-repos-from-pre-specified-gh-orgs

### DIFF
--- a/metashare/api/jobs.py
+++ b/metashare/api/jobs.py
@@ -200,10 +200,8 @@ def delete_scratch_org(scratch_org):
 delete_scratch_org_job = job(delete_scratch_org)
 
 
-def refresh_github_repositories_for_user_job(user):
+def refresh_github_repositories_for_user(user):
     user.refresh_repositories()
 
 
-refresh_github_repositories_for_user_job_job = job(
-    refresh_github_repositories_for_user_job
-)
+refresh_github_repositories_for_user_job = job(refresh_github_repositories_for_user)
--- a/metashare/api/jobs.py
+++ b/metashare/api/jobs.py
@@ -198,3 +198,12 @@ def delete_scratch_org(scratch_org):
 
 
 delete_scratch_org_job = job(delete_scratch_org)
+
+
+def refresh_github_repositories_for_user_job(user):
+    user.refresh_repositories()
+
+
+refresh_github_repositories_for_user_job_job = job(
+    refresh_github_repositories_for_user_job
+)
--- a/metashare/api/models.py
+++ b/metashare/api/models.py
@@ -37,17 +37,12 @@ class User(mixins.HashIdMixin, AbstractUser):
     objects = UserManager()
 
     def refresh_repositories(self):
-        self.notify_refreshing_repositories()
         repos = gh.get_all_org_repos(self)
         GitHubRepository.objects.filter(user=self).delete()
         GitHubRepository.objects.bulk_create(
             [GitHubRepository(user=self, url=repo) for repo in repos]
         )
         self.notify_repositories_updated()
-
-    def notify_refreshing_repositories(self):
-        message = {"type": "USER_REPOS_REFRESHING"}
-        async_to_sync(push.push_message_about_instance)(self, message)
 
     def notify_repositories_updated(self):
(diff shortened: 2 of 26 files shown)


wlonk:
@jgerigmeyer This requires a frontend change to handle the websocket event. Let me know if you wanna talk about that.

metashare/api/jobs.py
```diff
@@ -198,3 +198,10 @@ def delete_scratch_org(scratch_org):
 
 
 delete_scratch_org_job = job(delete_scratch_org)
+
+
+def refresh_github_repositories_for_user(user):
+    user.refresh_repositories()
```
jgerigmeyer: @wlonk This is also called from the `UserRefreshView` of `views.py` -- I assume that should also be converted to use the job instead.

wlonk: Thanks, yes, will do.

metashare/api/models.py
```diff
@@ -42,6 +42,17 @@ def refresh_repositories(self):
         GitHubRepository.objects.bulk_create(
             [GitHubRepository(user=self, url=repo) for repo in repos]
         )
+        self.notify_repositories_updated()
+
+    def notify_repositories_updated(self):
+        from .serializers import RepositorySerializer
+
+        gh_repositories = self.repositories.values_list("url", flat=True)
+        repositories = Repository.objects.filter(repo_url__in=gh_repositories)
```
jgerigmeyer: @wlonk It looks like this entirely bypasses pagination. It also seems a bit odd to me that we're duplicating similar logic here and in the `RepositoryViewSet`. I'd propose either returning no content here and allowing the front end to query the `RepositoryViewSet` endpoint, or making this `payload` actually use the `RepositoryViewSet` (including pagination).

jgerigmeyer: For now, we'll remove the payload from the `USER_REPOS_REFRESH` event and allow the front-end to trigger a new `GET`.

metashare/api/models.py
```diff
@@ -342,27 +352,29 @@ def notify_has_url(self):
...
+        message = {"type": "SCRATCH_ORG_DELETE", "payload": payload}
         async_to_sync(push.push_message_about_instance)(self, message)
 
 
 @receiver(user_logged_in)
 def user_logged_in_handler(sender, *, user, **kwargs):
-    user.refresh_repositories()
+    from .jobs import refresh_github_repositories_for_user_job
+
+    refresh_github_repositories_for_user_job.delay(user)
```
jgerigmeyer: Problem: On initial page-load, the front end has no way of knowing whether or not the user just logged in, and therefore whether or not there's a current refresh happening in the background. There's a delay of a few seconds before the job starts, and this handler fires too early to be useful (i.e. before the front end subscribes to the user WS channel).

Proposal: In addition to triggering an automatic refresh after every login, also trigger an automatic refresh on every page load _if the user has no repos_. That way on every page load, if there are no repos the front end can assume there's a background refresh happening, and wait for it to finish before displaying the app.

jgerigmeyer: @wlonk (cc @davisagli) I'd be happy for your approval or feedback on the above -- or to discuss it further on a call.

jgerigmeyer:
[Only query GitHub repos from pre-specified GH orgs.](https://trello.com/c/NfO93RKQ/64-only-query-github-repos-from-pre-specified-gh-orgs)

davisagli:
The "Only query GitHub repos from pre-specified GH orgs" part is the crucial thing I'm waiting for. My github user has access to a LOT of repositories.

jgerigmeyer:
> The "Only query GitHub repos from pre-specified GH orgs" part is the crucial thing I'm waiting for. My github user has access to a LOT of repositories.

@davisagli That makes sense... but I also think this change should allow you to test the app now, even if it takes awhile on login?

davisagli:
@jgerigmeyer yeah it'll certainly help. I'll give it a try on Monday.

wlonk:
I guess I'll do the "only query selected GH repos" in another branch then 😄

PR #182, opened 2019-09-20
Change Github OAuth scope to be configurable via env var

jlantz:
Also, sets defaults to only public repositories and not requesting access to Github orgs.


jgerigmeyer:
[I want to change the scope of the GH OAuth app per-deployment.](https://trello.com/c/DT7Lee21/89-i-want-to-change-the-scope-of-the-gh-oauth-app-per-deployment)

config/settings/base.py
```diff
@@ -295,7 +299,13 @@ def env(name, default=NoDefaultValue, type_=str):
 # WHITENOISE_ROOT = PROJECT_ROOT.joinpath(static_dir_root)
 
 SOCIALACCOUNT_PROVIDERS = {
-    "github": {"SCOPE": ["read:user", "user:email", "repo", "read:org"]},
+    "github": {
+        "SCOPE": env(
+            "GITHUB_OAUTH_SCOPES",
+            default="read:user,user:email,public_repo",
```
jgerigmeyer: Without the `read:org` scope, this currently throws an exception:

```
ForbiddenError: 403 You need at least read:org scope or user scope to list your organizations.
```

So if that's actually a use-case we want to support (that is, only showing public repos owned directly by the logged-in-user), we need to make some changes to `get_all_org_repos` in `gh.py` to handle that.

jlantz: Is it possible we could be using this API call instead?
https://developer.github.com/v3/repos/#list-your-repositories

Ideally, we shouldn't need users of MetaShare to grant access to every Github organization they're a member of.  We just need to know which repositories they're authorized to contribute to.

jgerigmeyer: D'oh -- yes, that works. Early on we were only seeing user-owned repos until we added in an `organizations` query, but apparently that had to do with other bugs which we since fixed. This works fine if we remove the specific query of the user's orgs. 👍

wlonk:
@jlantz what would you think about _adding_ that parsed env var to a minimal list of scopes we require for MetaShare to even work? À la `"SCOPE": ["read"] + env("GITHUB_SCOPES"...`?

jgerigmeyer:
> @jlantz what would you think about _adding_ that parsed env var to a minimal list of scopes we require for MetaShare to even work? À la `"SCOPE": ["read"] + env("GITHUB_SCOPES"...`?

@wlonk I like that idea -- basically requiring the basics that allow MS to work, but allowing additional scopes via env var.

jlantz:
@wlonk I'm wondering if perhaps it makes sense to only do a boolean env var instead called something like GITHUB_OAUTH_PRIVATE_REPO which if true would use `repo` instead of `public_repo` (the default).  There's not really much else that we'd want to make configurable

PR #220, opened 2019-10-03
[change-ws-events] Clean up websocket events

wlonk:
gecko

Let's try this out and see if it works!

### DIFF
--- a/metashare/api/jobs.py
+++ b/metashare/api/jobs.py
@@ -178,21 +178,21 @@ def get_unsaved_changes(scratch_org):
     try:
         old_revision_numbers = scratch_org.latest_revision_numbers
         new_revision_numbers = sf_changes.get_latest_revision_numbers(scratch_org)
-        unsaved_changes = sf_changes.compare_revisions(
+        scratch_org.unsaved_changes = sf_changes.compare_revisions(
             old_revision_numbers, new_revision_numbers
         )
-        if unsaved_changes:
-            scratch_org.latest_revision_numbers = new_revision_numbers
-            scratch_org.unsaved_changes = unsaved_changes
+        scratch_org.latest_revision_numbers = new_revision_numbers
     except Exception as e:
         scratch_org.unsaved_changes = {}
+        scratch_org.currently_refreshing_changes = False
+        scratch_org.save()
         async_to_sync(report_scratch_org_error)(
             scratch_org, e, "SCRATCH_ORG_FETCH_CHANGES_FAILED"
         )
         tb = traceback.format_exc()
         logger.error(tb)
         raise
-    finally:
+    else:
         scratch_org.currently_refreshing_changes = False
         scratch_org.finalize_get_unsaved_changes()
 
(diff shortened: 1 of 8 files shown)


metashare/api/jobs.py
```diff
@@ -247,21 +184,27 @@ def get_unsaved_changes(scratch_org):
...
+    except Exception as e:
+        scratch_org.unsaved_changes = {}
+        async_to_sync(report_scratch_org_error)(
+            scratch_org, e, "SCRATCH_ORG_FETCH_CHANGES_FAILED"
+        )
+        tb = traceback.format_exc()
+        logger.error(tb)
+        raise
+    finally:
+        scratch_org.currently_refreshing_changes = False
```
jgerigmeyer: Would it be clearer to move this into `finalize_get_unsaved_changes`, so that both changes to `currently_refreshing_changes` live together in the model?

metashare/api/jobs.py
```diff
@@ -284,15 +227,16 @@ def commit_changes_from_org(scratch_org, user, desired_changes, commit_message):
...
-    )
+    except Exception as e:
+        async_to_sync(report_scratch_org_error)(
+            scratch_org, e, "SCRATCH_ORG_COMMIT_CHANGES_FAILED"
+        )
+        tb = traceback.format_exc()
+        logger.error(tb)
+        raise
+    finally:
+        scratch_org.currently_capturing_changes = False
```
jgerigmeyer: ...and same question here.

metashare/api/jobs.py
```diff
@@ -284,15 +227,16 @@ def commit_changes_from_org(scratch_org, user, desired_changes, commit_message):
...
-        },
-    )
+    except Exception as e:
+        async_to_sync(report_scratch_org_error)(
+            scratch_org, e, "SCRATCH_ORG_COMMIT_CHANGES_FAILED"
+        )
+        tb = traceback.format_exc()
+        logger.error(tb)
+        raise
+    finally:
```
jgerigmeyer: Hm, using `finally` here (and above in `get_unsaved_changes`) means in an error case we get both the error message (e.g. `SCRATCH_ORG_COMMIT_CHANGES_FAILED`) _and_ the success message (e.g. `GITHUB_CHANGES_COMMITTED`). I'd propose changing it to an `else`, and then adding to the `except` block:

```py
scratch_org.currently_capturing_changes = False
scratch_org.save()
```

metashare/api/jobs.py
```diff
@@ -247,21 +184,27 @@ def get_unsaved_changes(scratch_org):
         if unsaved_changes:
             scratch_org.latest_revision_numbers = new_revision_numbers
             scratch_org.unsaved_changes = unsaved_changes
```
jgerigmeyer: I just noticed this now... but why are we only doing this `if unsaved_changes`? It means that if we think there are changes, then we refresh and discover there aren't any... the org doesn't get updated to reflect that there actually aren't changes. 🤔

metashare/api/jobs.py
wlonk: This is the scenario where I would want `finalize_*` to handle an error message I think.

jgerigmeyer: 👍 Yeah, I considered that. Maybe also moving `report_scratch_org_error` to the models as well.

PR #221, opened 2019-10-04
[limited-access-user] Fix devhub connection for limited access users

davisagli:
(as long as they have the necessary permissions to create scratch orgs)


jgerigmeyer:
[If there was an error or insufficient permissions connecting to a SF org, I want it to be exposed to the user.](https://trello.com/c/e1dkbOKV/73-if-there-was-an-error-or-insufficient-permissions-connecting-to-a-sf-org-i-want-it-to-be-exposed-to-the-user)

PR #222, opened 2019-10-04
[spinner-z-index] z-index is higher than modals and dropdowns

duggiemitchell:
See: https://trello.com/c/u7CKHXhl
bug
I found that setting z-index to 0 sets it to the "default layer" and seemed to work (there's no slds  design token for a 0 z-index).  

Also, added `<SpinnerWrapper/>` component so we don't have to pass the custom class to the spinner container, should we need to use it again.


jgerigmeyer:
[Bug: Spinners have higher z-index than dropdowns, modals.](https://trello.com/c/u7CKHXhl/95-bug-spinners-have-higher-z-index-than-dropdowns-modals)

PR #235, opened 2019-10-07
📦 Clean up deployment, unused deps, docker.

jgerigmeyer:
dog


docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
       DJANGO_SETTINGS_MODULE: config.settings.local
       REDIS_URL: redis://redis:6379
     volumes:
-      - .:/app
+      - .:/app:cached
```
jgerigmeyer: @wlonk I'm not sure I fully understand these caching settings (https://docs.docker.com/docker-for-mac/osxfs-caching/), but I read that this is a way to _significantly_ improve Docker performance on OS X. See https://github.com/docker/for-mac/issues/3499#issuecomment-538681270

docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
     environment:
       - POSTGRES_USER=metashare
     volumes:
-      - ./postgres:/var/lib/postgresql/data
+      - ./postgres:/var/lib/postgresql/data:cached
```
jgerigmeyer: I dunno if I'm using this too aggressively on too many volumes? 🤷‍♂

docker-compose.yml
wlonk: Read through those docs, and looks good. Let's try it out and keep an eye out for any problems.

PR #237, opened 2019-10-09
[clean-repository-handling] Allow GitHub with or without www. prefix

wlonk:
See https://trello.com/c/fsDdnJmD/101-chore-clean-up-repository-url-handling-in-admin

hiding rat

### DIFF
--- a/metashare/api/gh.py
+++ b/metashare/api/gh.py
@@ -55,14 +55,15 @@ def get_all_org_repos(user):
     return repos
 
 
-def normalize_github_url(url):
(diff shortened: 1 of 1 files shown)


wlonk:
@jgerigmeyer Yeah, I think I do like that better. It'll be a bit bigger (db migration etc) but I'll do it here.

davisagli:
@wlonk @jgerigmeyer for some other projects we've used the github api to look up github's integer id for the repo. that will be consistent even if the repo gets renamed or transferred to a different owner.

PR #241, opened 2019-10-14
Allow optional strings in settings

wlonk:
opossum


davisagli:
Closing since we went with https://github.com/oddbird/MetaShare/pull/32 instead, but we can revisit it.

PR #291, opened 2019-12-27
Merge from OddBird fork.

jgerigmeyer:
alpaca

- Submit projects for review [oddbird/MetaShare/pull/99]
- Do not log errors on GH webhook for `tag` push events [oddbird/MetaShare/pull/126]


jgerigmeyer:
@davisagli Deploying this will quiet down the Sentry errors for GH webhook push events for tags.

PR #305, opened 2020-03-30
pass through PATH to the subprocess so we can run sfdx

davisagli:
The subprocess change broke creating scratch orgs for sfdx-format projects, and I suspect this is the solution.


davisagli:
wrong target

PR #311, opened 2020-04-17
📦 update cumulusci etc

davisagli:


davisagli:
@jgerigmeyer what do I need to do to unbreak the yarn install? I wasn't intending to touch frontend stuff at all here.

jgerigmeyer:
@davisagli Ah, it's pulling the `latest` version of the Dockerfile, which we've updated. We've since used tagged versions, but that's not merged upstream yet. I'd think you could update the versions in `package.json`? But I'm okay just merging this and letting CI run on our `master`.

jgerigmeyer:
@davisagli Should this be based on https://github.com/oddbird/Metecho/ instead?

davisagli:
@jgerigmeyer I branched it from there, but submitted the PR to the wrong place. Will try again.

PR #312, opened 2020-04-22
Security enhancements

davisagli:
Following up on some requests from our security folks:
1. Restrict which Django management commands can be used in production
2. Don't use http:// in test data

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -2,8 +2,18 @@
 import os
 import sys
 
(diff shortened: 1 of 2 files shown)


davisagli:
@wlonk this was targeting the wrong master; replaced by https://github.com/oddbird/Metecho/pull/276

wlonk:
I spotted that as soon as I posted this and closed this tab and went to the next review 😆

PR #314, opened 2020-05-13
Handle scratch org features as a single string

davisagli:
This is supported by sfdx so we probably should too.


wlonk:
So I guess you don't need review on this one 😆

PR #711, opened 2021-04-01
Merge from OddBird fork.

jgerigmeyer:
- Allow creation of Project-level scratch orgs (https://github.com/oddbird/Metecho/pull/940/)
- Allow creation of Epic-level scratch orgs (https://github.com/oddbird/Metecho/pull/685/)
- Allow creation of Task-level scratch orgs (https://github.com/oddbird/Metecho/pull/1018)
- Add Project intro modal and "Plan" guided tour (https://github.com/oddbird/Metecho/pull/772/)
  - This can be disabled globally by setting env var `ENABLE_WALKTHROUGHS=False`
- Allow assigning GitHub users directly to Task roles (https://github.com/oddbird/Metecho/pull/884/)
- Confirm before deleting scratch org with unretrieved changes (https://github.com/oddbird/Metecho/pull/1015)
- Add "Test" task status in tasks table (https://github.com/oddbird/Metecho/pull/1017)
- Style Django-served 404 page (https://github.com/oddbird/Metecho/pull/1016)
- Upgrade dependencies
- Continue adding Storybook component library (not yet deployed publicly)


.gitignore
```diff
@@ -75,12 +73,6 @@ jscache/
 # Sphinx documentation
 docs/_build/
 
-# pyenv
-.python-version
```
davisagli, line 79: Why did you remove this? I use pyenv.

.python-version
```diff
@@ -0,0 +1 @@
+3.9.0
```
davisagli, line 1: This file should not be in source control. I have mine configured to use a virtualenv named `metecho`

.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+yarn-path ".yarn/releases/yarn-1.22.5.cjs"
```
davisagli, line 5: What is this?

config/settings/base.py
```diff
@@ -509,6 +511,8 @@ def env(name, default=NoDefaultValue, type_=str):
 # New feature branch prefix:
 BRANCH_PREFIX = env("BRANCH_PREFIX", default=None)
 
+ENABLE_WALKTHROUGHS = env("ENABLE_WALKTHROUGHS", default=True, type_=boolish)
```
davisagli, line 514: Don't bother changing it, but when we're putting a new feature behind a feature flag, it should probably be off by default.

.python-version
jgerigmeyer: This should only impact devs using pyenv, but I'll remove it.

.yarnrc
jgerigmeyer: The correct version of yarn is bundled with the repo (in `.yarn/releases`), as per yarn recommendations. Again, I can remove if you feel like that's an issue.

config/settings/base.py
jgerigmeyer: Makes sense. I went this way because I'm not used to using feature flags in this project and I was thinking of it as a temp workaround while we flesh out the other walkthroughs -- but I think you're right. 👍

.yarnrc
davisagli: Where can I read more about that recommendation? My initial reaction is that I don't love adding a large file to the repository if we don't need to, but I'd like to understand the argument in favor before deciding.

jgerigmeyer: Yeah, the argument is basically "so users don't need to worry about local yarn versions". https://classic.yarnpkg.com/en/docs/cli/policies

davisagli: It makes some sense; let's give it a try.

davisagli (approved):

PR #853, opened 2021-05-17
Handle any exception while checking if org is still accessible

davisagli:
Metecho uses this function to check if an org which may have been deleted is still accessible. cumulusci now has its own exception handling for the HTTPError when it is not, which means that this was no longer catching the right exception, so we showed an error to the user where we shouldn't. Let's catch any error while trying to get an access token so that we don't do that.


Br4nd0R (approved):

jgerigmeyer:
[Bug: Metecho no longer knows when a SF org has expired.](https://trello.com/c/3WR5iRoC/282-bug-metecho-no-longer-knows-when-a-sf-org-has-expired)

PR #883, opened 2021-05-20
handle more exceptions indicating org can't be accessed

davisagli:
This fix is similar to #853. CumulusCI now handles the HTTPError that occurs if the org can no longer be authenticated (i.e. because it expired) and raises a different exception.


jlantz (approved):

jgerigmeyer:
[Bug: Metecho no longer knows when a SF org has expired.](https://trello.com/c/3WR5iRoC/282-bug-metecho-no-longer-knows-when-a-sf-org-has-expired)

PR #1249, opened 2021-09-08
[Metecho] CTC integration

jofsky:


.heroku/check_change_traffic_control.py
```diff
@@ -0,0 +1,85 @@
...
+import jwt
+import pydantic
+import requests
+
+logger = logging.getLogger()
+
+
+class CTCSettings(pydantic.BaseSettings):
+    metaci_release_webhook_url: str
+    metaci_release_webhook_auth_key: str
```
davisagli, line 16: Use ctc_url and ctc_auth_key here; this isn't MetaCI

davisagli (changes requested):

davisagli (approved):

PR #2019, opened 2022-06-08
Allow creating new Projects (from OddBird).

jgerigmeyer:
~~**NOTE: This should be reviewed and merged after https://github.com/SFDO-Tooling/Metecho/pull/2018 is already merged.**~~

Features:
- Allow creating new Projects/Repos (https://github.com/oddbird/Metecho/pull/1376)


src/js/components/projects/createProjectModal/summary.tsx
```diff
@@ -0,0 +1,69 @@
...
+  name,
+  selectedOrg,
+  selectedCollaborators,
+  selectedDeps,
+}: Props) => {
+  const { t } = useTranslation();
+
+  return (
+    <form className="slds-form slds-p-around_large">
+      <h3 className="slds-text-heading_small slds-m-bottom_medium">
```
davidjray, line 27: this component seems to have a lot of H3s. Is this a valid heading structure? It is important that headers follow a logical hierarchy

src/js/components/projects/listItem.tsx
```diff
@@ -1,75 +1,79 @@
...
+        </div>
+        <div
+          className="slds-media__body
+            slds-border_left
+            slds-p-around_small
+            slds-grid
+            slds-grid_vertical
+            content-fill-height"
+        >
+          <h2
```
davidjray, line 56: Speaking of header ordering, it appears the list component is using an H2 tag, but then the listItems are also using an H2. The heading order here should be taken into consideration.

davidjray (changes requested):
Overall from what I can tell this looks good, I just have some concerns regarding header ordering.

metecho/api/constants.py
```diff
@@ -1,3 +1,19 @@
+from django.utils.translation import gettext as _
+
 CHANNELS_GROUP_NAME = "{model}.{id}"
 ORGANIZATION_DETAILS = "organization_details"
 LIST = "list"
+
+
+class GitHubAppErrors:
```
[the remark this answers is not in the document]
jgerigmeyer: Yes, that's my understanding. @jerivas Is that right?

src/js/components/projects/createProjectModal/summary.tsx
jgerigmeyer: Hm, my understanding (from e.g. [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements)) is that multiple H3s is a valid structure for repeated headings of a similar level.

Since SLDS [updated the Modal heading to an H1](https://github.com/salesforce/design-system-react/releases/tag/v0.10.48), I think these should probably be H2s though -- I'll update that (and other similar sub-headings inside Modals).

Adjusted in https://github.com/SFDO-Tooling/Metecho/pull/2019/commits/8b50b4fc5c793962be86b31a7706446e63cc8a49

src/js/components/projects/listItem.tsx
jgerigmeyer: Good catch! Fixed in https://github.com/SFDO-Tooling/Metecho/pull/2019/commits/8b50b4fc5c793962be86b31a7706446e63cc8a49

metecho/api/constants.py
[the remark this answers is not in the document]
jerivas: Yes, that was my intention

davidjray (approved):
Nice, thanks for making these changes!

jgerigmeyer:
@davidmreed Is this ready to be merged?

PR #2042, opened 2022-10-07
Support VBT-reliant flows in Metecho

prescod:
Metecho equivalent of: https://github.com/SFDO-Tooling/MetaDeploy/pull/3468

Make a [new script](https://github.com/SFDO-Tooling/Metecho/blob/363a17e96672a34ffa48681592c3c561fa498098/.heroku/start_worker.sh) that makes Chrome available to the VBT compiler.

Invoke it from [Procfile](https://github.com/SFDO-Tooling/Metecho/blob/363a17e96672a34ffa48681592c3c561fa498098/Procfile).

Install Vlocity and Upgrade "[MiniPass](https://github.com/SFDO-Tooling/Metecho/compare/feature/support-vbt?expand=1#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de)" to work around a bug that seems to have arisen asynchronously (perhaps the type declarations are downloaded at runtime).

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,4 +1,4 @@
 web: yarn django:serve:prod
-worker: python manage.py rqworker default
+worker: sh .heroku/start_metadeploy_worker.sh
 worker-short: honcho start -f Procfile_worker_short
 release: ./.heroku/release.sh
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
     "reselect": "^4.1.6",
     "sockette": "^2.0.6",
     "uuid-random": "^1.3.2",
+    "vlocity": "^1.15.6",
     "whatwg-fetch": "^3.6.2"
   },
   "devDependencies": {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -157,6 +158,8 @@
     "@storybook/**/glob-parent": "^5.1.2",
     "@storybook/**/trim-newlines": "^3.0.1",
     "i18next-scanner-webpack/**/glob-parent": "^5.1.2",
+    "//": "Related to  https://github.com/isaacs/minipass/pull/41 ",
```
jstvz, line 161: 😬

jstvz (approved):

prescod:
Todo:

1. make or find a test repo that includes components that need compilation
2. build a scratch org against the test repo and check that the components compiled
3. document the process
4. make a new WI about automating the process

For MetaDeploy, we were using this repo:
* https://github.com/Industries-SolutionFactory-Connect/cme-automation

But we should really have our own. Probably as an extension of CCI-Test or Food Bank.

Procfile
```diff
@@ -1,4 +1,4 @@
 web: yarn django:serve:prod
-worker: python manage.py rqworker default
+worker: sh .heroku/start_metadeploy_worker.sh
```
jofsky, line 2:
```suggestion
worker: sh .heroku/start_worker.sh
```

PR #2048, opened 2022-10-17
Path Component Accessible label

davidjray:
Adding accessible label to Path component

Note the test snapshots are auto-generated by adding the `--updateSnapshot` flag when running JS tests.


davidjray:
Epic screen example:
Screen Reader before change
Screen Shot 2022-10-18 at 10 09 12 AM

After Change
Screen Shot 2022-10-18 at 10 10 31 AM

locales_dev/en/translation.json
```diff
@@ -143,6 +143,7 @@
   "Epic Name": "Epic Name",
   "Epic Options": "Epic Options",
   "Epic Scratch Org": "Epic Scratch Org",
+  "Epic Status": "Epic Status",
```
davidmreed, line 146: I believe these new strings need to be copied over to `locales/en/translation.json` to be included in the next localization pass.

src/js/components/path/index.tsx
```diff
@@ -7,16 +7,18 @@ interface PathProps {
...
-const Path = ({ steps, activeIdx, isCompleted, isLost }: PathProps) => (
+const Path = ({ steps, activeIdx, isCompleted, isLost, label }: PathProps) => (
   <div className="slds-region_small">
     <div className="slds-path">
       <div className="slds-grid slds-path__track">
         <ul
           className="slds-path__nav"
           role="listbox"
           aria-orientation="horizontal"
+          aria-label={label}
```
davidmreed, line 21: Is an empty string a valid value for `aria-label`? Just noting that `label` is nullable - do we need to worry about that?

davidmreed (changes requested):

davidmreed (approved):

src/js/components/path/index.tsx
davidjray: Empty `aria-labels` are ignored, but great question!

PR #2057, opened 2022-10-26
Update to Heroku Stack 22

TheBitShepherd:
* Nothing for release notes

The version of python we were using (3.10.2) [isn't supported](https://devcenter.heroku.com/articles/python-support#supported-runtimes) on Heroku 22.


runtime.txt
```diff
@@ -1 +1 @@
-python-3.10.2
+python-3.10.8
```
jstvz, line 1: 💭 : I wish there was an automated way to keep `runtime.txt` updated.

jstvz (approved):

runtime.txt
TheBitShepherd: If stack 22 is supported through 2027—in theory—we shouldn't need to update this again for five years, right?

PR #2058, opened 2022-10-27
Heroku stack-22 changes

TheBitShepherd:
* Nothing for release notes

We now pass `PYTHONPATH`, `LD_LIBRARY_PATH`, and `LIBRARY_PATH` to the subprocess where flow are run in Metecho.

### DIFF
--- a/metecho/api/sf_run_flow.py
+++ b/metecho/api/sf_run_flow.py
@@ -390,6 +390,19 @@ def run_flow(*, cci, org_config, flow_name, project_path, user):
         "HOME": project_path,
         "PATH": os.environ["PATH"],
     }
+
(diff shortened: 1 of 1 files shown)


metecho/api/sf_run_flow.py
```diff
@@ -390,6 +390,19 @@ def run_flow(*, cci, org_config, flow_name, project_path, user):
         "HOME": project_path,
         "PATH": os.environ["PATH"],
     }
+
+    # Pass exta env vars when running in the context of Heroku stack-22.
```
davidmreed, line 394:
```suggestion
    # Pass extra env vars when running in the context of Heroku stack-22.
```

davidmreed (approved):

PR #2067, opened 2022-11-03
Prevent 403 when copying branch protection rules

TheBitShepherd:
This PR covers two separate bugs:

1. When branch protection rules _are not_ enabled on a branch, GitHub will return a 404. This exception was not being properly caught and handled, but now is.
2. When branch protection rules _are_ enabled we are running into an error where the GitHub API is returning a 403. This error is outlined in more detail in W-12000044.

We also now prompt the user to enable branch protection rules are a repository is successfully created:
Screen Shot 2022-11-03 at 10 58 06 AM

### DIFF
--- a/metecho/api/jobs.py
+++ b/metecho/api/jobs.py
@@ -38,7 +38,6 @@ from github3.repos.repo import Repository
 
 from .email_utils import get_user_facing_url
 from .gh import (
-    copy_branch_protection,
     get_all_org_repos,
     get_cached_user,
     get_cumulus_prefix,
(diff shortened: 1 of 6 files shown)


metecho/api/jobs.py
```diff
@@ -280,10 +279,11 @@ def create_repository(
...
         # Copy branch protection rules from the template repo
-        if tpl_repo:
-            copy_branch_protection(
-                source=tpl_repo.branch(branch_name), target=repo.branch(branch_name)
-            )
+        # See copy_branch_protection() for why we don't use this currently
+        # if tpl_repo:
+        #     copy_branch_protection(
+        #         source=tpl_repo.branch(branch_name), target=repo.branch(branch_name)
+        #     )
```
TheBitShepherd, line 286: I really dislike commented code, but as this seems to be an issue with GitHub App perms I'd like to keep this around. I'm considering a more concise repro and logging the issue with them.

metecho/api/jobs.py
```diff
@@ -875,9 +875,9 @@ def refresh_github_users(project: Project, *, originating_user_id):
         for collaborator in repo.collaborators():
             try:
                 # Retrieve additional information for each user by querying GitHub
-                collaborator.name = get_cached_user(
-                    GitHub(repo), collaborator.login
-                ).name
+                collaborator.name = (
+                    get_cached_user(GitHub(repo), collaborator.login).name or ""
+                )
```
TheBitShepherd, line 880: This is a fix to another error I was seeing where user's where trying to be created with null name values. The except block below attempts to mitigate one version of this, but I was seeing another.

davidmreed (approved):

PR #2071, opened 2022-11-09
Change heading to H2 to better fit page order

davidjray:
Simple single line heading fix, but all the test snapshots needed an update.

### DIFF
--- a/src/js/components/steps/index.tsx
+++ b/src/js/components/steps/index.tsx
@@ -14,7 +14,7 @@ const Steps = ({
   const someAssignees = steps.some((i) => i.assignee);
   return (
     <>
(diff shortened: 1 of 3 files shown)


boakley (commented):
This looks like a safe, simple change. However, the tests only seem to verify that the pages have an H3, they don't show that an a11y audit failed before the change and passes after. 

Do we even have a testing framework in place to run an audit and then verify specific failures don't appear?

davidjray:
@boakley we don't have a testing framework in place for accessibility issues yet. My short term goal is to remediate the outstanding bugs from the audit, then put something in place that will help ongoing development to include accessibility testing.

boakley (approved):
Looks good to me.

PR #2072, opened 2022-11-11
First attempt at solving a11y bug for a long loading message. 

boakley:
I haven't updated the tests yet.

I added 'role="status"' to two divs that include the message. There is a third block that did not wrap the message in a div so I did that. I chose to use the same pattern as the other blocks using a `React.Fragment` (`<>`) but I don't know if I should or not since it only contains a single div.

Rookie mistake or wise choice to be consistent? You be the judge.


davidjray (approved):
Looks great.

PR #2077, opened 2022-11-17
Override toast focus to pre-determined div

davidjray:
Almost all the toasts in Metecho are generated from some type of async process, like an error, or the creation of a scratch org etc. This means the element that creates the toast is almost never available, so the focus is sent to the `<body>` of the HTML. This could be confusing or inconvenient for users, so we instead move focus for the main div wrapping the meaningingful content of the page.

Control of focus aft4er a toast is closed is be controlled by the `<toasts>` component. Any main layout component that needs to take advantage of the focus adds `metecho-toast-focus` as a css class to a wrapping `<div>` with `tabIndex={-1}` which allows the div to become focusable.

Also I updated the markup of `<DetailPageLayout>` to include `<main>` and `<aside>` which are more appropriate semantic elements.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,30 +1,55 @@
-{% load api_bootstrap %}
-{% load js_reverse %}
-{% load static %}
+{% load api_bootstrap %} {% load js_reverse %} {% load static %}
 
 <!DOCTYPE html>
 <html lang="en" dir="ltr" class="slds-theme_default">
   <head>
     <meta charset="utf-8" />
     <meta http-equiv="x-ua-compatible" content="ie=edge" />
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -36,12 +61,23 @@
     <!-- JavaScript and Styles are inserted here -->
   </head>
   <body>
-    <main id="app" class="slds-grid slds-grid_vertical"{% if user.is_authenticated %} data-user="{{ user|serialize }}"{% endif %}>
+    <div
```
davidjray, line 64: This is the important change in this file, the rest is whitespace from the linter.

TheBitShepherd (approved):
This looks great, nice work @davidjray!

PR #2079, opened 2022-12-01
Remove Duplicate H1 from PageHeader

davidjray:
Removed the use of the `PageHeader` component, as our usage was forcing additional H1 elements into the markup. I swapped the `PageHeader` with some of the basic makeup for the visual display of the page header, while removing the H1 element embedded in the page header component.


boakley (commented):
This looks good.  It looks like you copied the markup created by `<PageHeader>` and just removed the `<h1>` tag.

boakley (approved):
looks good to me.

PR #2081, opened 2022-12-01
Update the state after retrieving data/omnistudio stuff

prescod:
W-11976440

Previously, you could not advance to "Submit changes for Testing" unless you retrieve Metadata changes.Now Data or Omnistudio changes will also allow you to advance.

### DIFF
--- a/metecho/api/jobs.py
+++ b/metecho/api/jobs.py
@@ -1363,6 +1363,9 @@ def commit_dataset_from_org(
                     branch=task.branch_name,
                     commit_message=commit_message,
                 )
+                org.task.has_unmerged_commits = True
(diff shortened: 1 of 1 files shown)


prescod:
@TheBitShepherd  https://github.com/SFDO-Tooling/Metecho/pull/2087

prescod:
> Nice fix. Should we add an `assert` to `metecho/api/tests/jobs.py::TestCommitDatasetFromOrg::test_ok` to prove that the `has_unmerged_commits` attribute was set properly?

@TheBitShepherd https://github.com/SFDO-Tooling/Metecho/pull/2087

PR #2085, opened 2022-12-08
Re-Fix toast focus management when closing toasts without breaking home-page

davidjray:
This PR re-appies previous focus management changes that were reverted due to a previous bug that broke the homepage. 

The prettier tool is not compatible with Django templates, so it formatted the templates in a way that prevented the app from loading when logged out.

<del>A new python test was added as a smoke test to ensure the home page template loads as expected, especially in the context of a logged out user.</del>

Additionally, an exclusion was added to the `.prettierignore` file so HTML based Django templates will not get formatted again in the future.


src/js/components/utils/detailPageLayout.tsx
```diff
@@ -147,12 +147,15 @@ const DetailPageLayout = ({
...
+        // tabIndex required for toast close focus
+        tabIndex={-1}
+        className="metecho-toast-focus
+          slds-p-around_x-large
           slds-grid
           slds-gutters
           slds-wrap"
       >
-        <div
+        <aside
```
boakley, line 158: Should this `<aside>` node also have a role, or are a11y roles implicit in this tag? I found [this page](https://www.w3.org/WAI/GL/wiki/Using_HTML5_aside_element) that says:

> adding the ARIA role="note" as follows or role="complementary" is advised.

src/js/components/utils/detailPageLayout.tsx
```diff
@@ -167,16 +170,16 @@ const DetailPageLayout = ({
             />
           )}
           {sidebar}
-        </div>
-        <div
+        </aside>
+        <main
```
boakley, line 174: Before today I didn't know about the `main` tag. This is a great learning moment for me. Making sure every page has one `main` tag seems like a best practice.

boakley (approved):
Looks good to me. I like the addition of the simple test for the login page. Simple but effective.

src/js/components/utils/detailPageLayout.tsx
davidjray: According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/complementary_role the "complimentary" landmark role is not needed, as it it implied with "<aside>".

davidjray: I'm also noticed on the page you linked it says "If support for HTML5 within a user agent is in doubt, then add the ARIA role="complementary"." I don't think we need to worry about HTML5 support.

src/js/components/utils/detailPageLayout.tsx
davidjray: Good semantic HTML markup is the most effective tool we have for accessibility. Here's a quick little article that covers the common semantic elements. 

https://uxdesign.cc/semantic-html-the-foundation-of-web-accessibility-e5bbecad7c17?gi=7398cc5a27d7

boakley (approved):
Looks good to me

PR #2098, opened 2023-02-28
Move daphne to prod dependencies

davidmreed:


prescod (approved):
Minor issue but we should probably remove it from dev.in too.

PR #2107, opened 2023-03-20
Download logs org failure

davidmreed:
When a scratch org build fails, allow the user to download the log from the Toast message, and offer them the org id to seek support.


src/js/store/orgs/actions.ts
```diff
@@ -182,11 +182,27 @@ export const provisionFailed =
...
           heading: msg,
-          details: message,
+          details: detailMsg,
           variant: 'error',
+          linkUrl: window.api_urls.scratch_org_log(model.id),
+          linkText: t('Download build log.'),
+          linkDownload: true,
+          linkDownloadFilename: t('Metecho Org {{orgId}} Build Log.txt', {
+            message,
+            orgId: model.id,
```
jstvz, line 204: What does `i18next.t` do when the key is missing? Fail silently?

docs/auth.md
```diff
@@ -0,0 +1,171 @@
+# Metecho Authentication
```
TheBitShepherd, line 1: This section is much need. Thanks for writing this up!

davidmreed: Oh jeez that was supposed to be in a different PR entirely. Retracting and pushing elsewhere...

jstvz (approved):
@davidmreed LGTM, but the code coverage threshold is 100% and it looks like you've got an uncovered branch somewhere blocking merge.

davidmreed:
W-12727028

prescod (approved):

PR #2113, opened 2023-04-11
Remove collaborator card heading

davidjray:
Swap default SLDS `<card>` react component with basic SLDS markup to remove an out of place H2. This is to fix an accessibility issue with confusing heading order.

In addition, I also fixed another potential issue, with each collaborator listed having the same title of "Remove", this would be confusing for users, so I made sure each button specifies which user would be deleted. This change required a simple update to the test.

### DIFF
--- a/test/js/components/epics/detail.test.jsx
+++ b/test/js/components/epics/detail.test.jsx
@@ -710,7 +710,11 @@ describe('<EpicDetail/>', () => {
           },
         },
       });
-      fireEvent.click(getByTitle('Remove'));
+      fireEvent.click(
+        getByTitle(
+          `Remove ${sampleGitHubUser2.name} (${sampleGitHubUser2.login})`,
+        ),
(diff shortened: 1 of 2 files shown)


test/js/components/epics/detail.test.jsx
```diff
@@ -710,7 +710,11 @@ describe('<EpicDetail/>', () => {
           },
         },
       });
-      fireEvent.click(getByTitle('Remove'));
+      fireEvent.click(
+        getByTitle(
+          `Remove ${sampleGitHubUser2.name} (${sampleGitHubUser2.login})`,
```
prescod, line 715: I don't see the stuff in parentheses in the title:

          `Remove ${sampleGitHubUser2.name} (${sampleGitHubUser2.login})`,

Why?

prescod (approved):

PR #2124, opened 2023-05-25
Allow SiteProfile to specify a default project namespace

jlantz:
Add the project_namespace field to SiteProfile and use the value if provided to set the CumulusCI project's namespace on new project creation.


davidmreed:
@jlantz Can you share some context on this proposal? It seems unusual to me that every project in a tenant would be expected to have (a | the same) namespace.

Do we need to consider improving the project creation UI to offer more interactivity?

PR #2154, opened 2023-12-11
@W-14603959 - Update backend deps and test failure fixes

vsbharath:


metecho/api/models.py
```diff
@@ -716,8 +716,9 @@ def __str__(self):
         return self.name
 
     def save(self, *args, **kwargs):
+        obj = super().save(*args, **kwargs)
         self.update_status()
-        return super().save(*args, **kwargs)
+        return obj
```
jstvz, line 722: [Django 4.1 introduces a change](https://docs.djangoproject.com/en/5.0/releases/4.1/#reverse-foreign-key-changes-for-unsaved-model-instances) where using reverse foreign keys before saving an object raises a `ValueError`. Since the `update_status` method interacts with related objects, as seen in methods like `should_update_review`, we have issues with reverse foreign key lookups.

You're correct that calling `super().save()` to ensure the `Epic` instance has an ID before `update_status` is the right fix. But the way it's written now, the status gets updated but not saved in the database.

```python
    def save(self, *args, **kwargs):
        if not self.id:
            super().save(*args, **kwargs)
        self.update_status()
        return super().save()

```

This code makes sure that if the `Epic` is new, we save it before updating the status. By not repeating `*args, **kwargs` in the final save, we avoid duplicating related objects.

requirements/prod.txt
```diff
@@ -4,100 +4,85 @@
...
     # via
     #   cumulusci
     #   python3-openid
-deprecated==1.2.13
-    # via redis
-dj-database-url==0.5.0
+dj-database-url==2.1.0
     # via -r requirements/prod.in
-django==4.0.6
+django==5.0
```
jstvz, line 81: Let's stick to the LTS for now. You can restrict which version of Django we use in `prod.in`:

```
Django~=4.2.0 # LTS
```

(Where `~=` means [compatible release](https://peps.python.org/pep-0440/#compatible-release).)

metecho/oauth2/salesforce/tests/provider.py
```diff
@@ -1,14 +1,29 @@
...
     assert "prompt" in result and result["prompt"] == "login"
 
 
+@pytest.mark.django_db
 def test_extract_uid(rf):
     request = rf.get("/")
-    provider = CustomSalesforceProvider(request)
+    provider = CustomSalesforceProvider(request, dummy)
     result = provider.extract_uid({"organization_id": "ORG", "user_id": "USER"})
     assert result == "ORG/USER"
```
jstvz, line 29: My mistake. Let's just use the factory_boy fixture instead:

```python
@pytest.mark.django_db
def test_get_auth_params(rf, social_app_factory):
    request = rf.get("/")
    app = social_app_factory(
        name=CustomSalesforceProvider.id,
        provider=CustomSalesforceProvider.id,
    )
    result = CustomSalesforceProvider(request, app).get_auth_params(request, None)
    assert "prompt" in result and result["prompt"] == "login"
```

metecho/api/tests/admin.py
```diff
@@ -42,7 +42,7 @@ def test_queryset__deleted(self, epic_factory):
 
         args = (
             None,
-            {"deleted_at": "true"},
+            {"deleted_at": ["true"]},
```
jstvz, line 45: We can set this aside until we do move to Django 5x:
```suggestion
            {"deleted_at": "true"},
```

jstvz (changes requested):
Thanks for slogging through these. Some changes requested, including sticking to Django 4.2.x. (⚠️: You may hit pennersr/django-allauth#3564 if you also update `django-allauth` to 0.59.0)

vsbharath:
@jstvz Thanks for the detailed review. Addressed the review comments and got to know the usage of factory_boy now.

jstvz (approved):

PR #2160, opened 2024-02-06
@W-14870747 - Dependecy updates and disabling Socalapp in db

vsbharath:


lakshmi2506 (approved):
LGTM

metecho/oauth2/adapter.py
```diff
@@ -9,4 +9,4 @@ class CustomSocialAccountAdapter(DefaultSocialAccountAdapter):
     def authentication_error(self, *args, **kwargs):
```
jstvz, line 9: Tests are still passing, but this [has been deprecated](https://github.com/pennersr/django-allauth/blob/079872ab5564fa73a73bbf36af990a29dea1ac3c/allauth/socialaccount/adapter.py#L72-L75). It'll need to be changed to `on_authentication_error` here and in the tests.

vsbharath: Yes, i will make this consistent with on_authentication_error. I just updated only when calling super().on_authentication_error(*args, **kwargs)

jstvz (approved):

